Issue #44, opened 2022-12-14
 Revise the application example implementation


acezen:
**Is your feature request related to a problem? Please describe.**
Currently the [examples]() of GraphAr are implement like unit test and they are not intuitive for user or developer beginner to know how to use GraphAr as example.
We need to revise the implement and make them more like an example and show case.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
related to #37

lixueclaire:
solved by #46

PR #46, opened 2022-12-16
organize an example pagerank app employing the gar library (#44)

andydiwenzhu:
## Proposed changes

Organize the original PageRank test code as an example application. It demonstrates how to edit CMake to integrate GAR C++ library and how to extend an input graph with its PageRank values using GAR methods.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [x] Documentation Update (if none of the other choices apply)

## Checklist

- [x] I have signed the CLA

## Further comments

### DIFF
--- a/examples/pagerank/CMakeLists.txt
+++ b/examples/pagerank/CMakeLists.txt
@@ -1,0 +1,23 @@
+cmake_minimum_required(VERSION 2.8)
+
+# Avoid warning about DOWNLOAD_EXTRACT_TIMESTAMP in CMake 3.24:
+if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
+  cmake_policy(SET CMP0135 NEW)
+endif()
+
+project(Pagerank)
+
+find_package(Threads REQUIRED)
+find_package(gar REQUIRED)
+
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+include(apache-arrow)
+build_arrow()
+
+include_directories(BEFORE SYSTEM ${ARROW_INCLUDE_DIR})
+
+include_directories(${GAR_INCLUDE_DIRS})
+
+add_executable(pagerank pagerank.cc)
+target_compile_features(pagerank PRIVATE cxx_std_17)
+target_link_libraries(pagerank PRIVATE ${GAR_LIBRARIES})
--- a/examples/pagerank/cmake/apache-arrow.cmake
+++ b/examples/pagerank/cmake/apache-arrow.cmake
@@ -1,0 +1,92 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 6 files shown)


examples/pagerank/pagerank.cc
```diff
@@ -0,0 +1,116 @@
...
+
+#include "gar/graph.h"
+#include "gar/graph_info.h"
+#include "gar/reader/arrow_chunk_reader.h"
+#include "gar/writer/arrow_chunk_writer.h"
+
+
+int main(int argc, char* argv[]) {
+  // read file and construct graph info
+  std::string path = "/tmp/ldbc_sample/parquet/ldbc_sample.graph.yml";
```
acezen, line 29: This is a little hard-code. Use the relative path to `test/gar-test` or add a soft link to it maybe better

examples/pagerank/pagerank.cc
```diff
@@ -0,0 +1,116 @@
...
+      if (out_degree[i] == 0)
+        pr_next[i] += damping * pr_curr[i];
+      pr_curr[i] = pr_next[i];
+      pr_next[i] = 0;
+    }
+  }
+
+  // extend the original vertex info and write results to gar using writer
+  // construct property group
+  GraphArchive::Property pagerank = {
```
acezen, line 82: As a pagerank example, I don't know if we just write the result to file not extend as a property would be better. That would have two benefits:
- simpler and clear for a example
- no need to include arrow header any more.
@lixueclaire what do you think?

examples/pagerank/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+cmake_minimum_required(VERSION 2.8)
```
acezen, line 1: I'm not sure if we need a CMakeList here. Seems put these to CMakeLists in root path and make example as a new target would be better. 
A new CMakeList.txt here would bring some cons:
- it need user/develop to **install** GraphAr first, not just make
- it's building need to build arrow again.

examples/pagerank/cmake/apache-arrow.cmake
```diff
@@ -0,0 +1,92 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
acezen, line 1: Ditto. We can reuse the apache-arrow.cmake of GraphAr.

acezen (commented):
Nice work! The pull request looks good. It has basically met the issue requirement but some modification still need.

examples/pagerank/CMakeLists.txt
andydiwenzhu: Yes. I think what you suggested is the in-project examples. What I want to show here is a new application employing the GAR library, in which case, users should first install GAR, and then build their own application projects.

examples/pagerank/pagerank.cc
andydiwenzhu: Agree.

examples/pagerank/pagerank.cc
andydiwenzhu: My purpose is to make this example a new project over the GAR library.

examples/pagerank/cmake/apache-arrow.cmake
andydiwenzhu: I prefer remove arrow from our example

examples/bfs_father_example.cc
```diff
@@ -6,7 +6,7 @@ You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
+Unless assertd by applicable law or agreed to in writing, software
```
acezen, line 9: Sorry, Why we need to change `required` to `assertd` here and the `assertd` has a typo ?

Issue #130, opened 2023-04-03
Provide a method to get the number of edges of the EdgesCollection


lixueclaire:
**Is your feature request related to a problem? Please describe.**
The current high-level EdgeIter only supports ++ operation, and cannot support random access of the edge. This makes it necessary to have O(m) time complexity to obtain the total number of edges through the EdgeIter, which is inefficient. And getting the number of edges is frequently to be used in real cases.
Therefore, it is suggested to add a method to the code that can directly obtain the number of edges in the graph.

**Describe the solution you'd like**
For ordered_by_source/ordered_by_dest adjList types, we can calculate the total edge number by checking the offset table for all vertex chunks. But for unordered edges, there are no offset tables. So maybe we need to create the edge count files (similar to the vertex_count file) to record the total number of edges for each vertex chunk.

**Additional context**
related to issue #78

lixueclaire:
This issue is discussed in GraphAr Weekly Community Meeting [Tuesday, April 4, 2023]:

lixueclaire:
solved by #132 & #133

Issue #78, opened 2023-01-11
Improve the performance of high-level graph iterators of the C++ library


lixueclaire:
**Is your feature request related to a problem? Please describe.**
An important application case of GraphAr is to serve out-of-core graph processing scenarios. With the graph data saved as GAR files in the disk, GraphAr provides a set of reading interfaces to allow to load part of graph data into memory when needed, to conduct analytics. 
Since for out-of-core graph processing, disk I/O time usually dominates the overall execution time. It is critically important that the GraphAr C++ library perform efficiently for traversing vertices/edges through high-level graph iterators.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

lixueclaire:
Currently, creating or copying an iterator (especially the EdgeIter) is time-consuming, since it holds all necessary readers (considering sharing readers is not thread-safe). Is there any way to let the iterator to be more light-weight?

lixueclaire:
> Currently, creating or copying an iterator (especially the EdgeIter) is time-consuming, since it holds all necessary readers (considering sharing readers is not thread-safe). Is there any way to let the iterator to be more light-weight?

This issue is discussed in GraphAr Weekly Community Meeting [Tuesday, April 4, 2023]: 
Perhaps we could improve the performance of EdgeIter by moving the readers to EdgesCollection, making the iterator more lightweight. Additionally, we could maintain a pool of readers and associate them with each iterator, so that when needed, the iterator can retrieve the appropriate reader from the pool. This approach would ensure thread-safety.

PR #133, opened 2023-04-11
[Improvement][C++] Improve the usability of EdgesCollection

lixueclaire:
## Proposed changes

This PR improves the usability of EdgesCollection by: 
- support to construct the EdgesCollection for a range of vertex chunks
- support to get the number of edges in a EdgesCollection

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

related to issue #130 & issue #78

### DIFF
--- a/cpp/include/gar/graph.h
+++ b/cpp/include/gar/graph.h
@@ -628,39 +628,6 @@ class EdgesCollection<AdjListType::ordered_by_source> {
   static const AdjListType adj_list_type_;
 
   /**
-   * @brief Initialize the EdgesCollection.
-   *
-   * @param edge_info The edge info that describes the edge type.
-   * @param prefix The absolute prefix.
-   */
-  EdgesCollection(const EdgeInfo& edge_info, const std::string& prefix)
-      : edge_info_(edge_info), prefix_(prefix), chunk_begin_(0) {
-    std::string base_dir;
-    GAR_ASSIGN_OR_RAISE_ERROR(auto fs,
-                              FileSystemFromUriOrPath(prefix, &base_dir));
-    GAR_ASSIGN_OR_RAISE_ERROR(auto adj_list_path_prefix,
-                              edge_info.GetAdjListPathPrefix(adj_list_type_));
-    base_dir += adj_list_path_prefix;
-    GAR_ASSIGN_OR_RAISE_ERROR(auto vertex_chunk_num,
-                              fs->GetFileNumOfDir(base_dir));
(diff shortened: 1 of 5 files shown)


cpp/include/gar/graph.h
```diff
@@ -656,32 +645,32 @@ class EdgesCollection<AdjListType::ordered_by_source> {
...
+   * @param vertex_chunk_begin The index of the begin vertex chunk.
+   * @param vertex_chunk_end The index of the end vertex chunk (not included).
    */
   EdgesCollection(const EdgeInfo& edge_info, const std::string& prefix,
-                  IdType chunk_begin, IdType chunk_end)
-      : edge_info_(edge_info),
-        prefix_(prefix),
-        chunk_begin_(chunk_begin),
-        chunk_end_(chunk_end) {
+                  IdType vertex_chunk_begin, IdType vertex_chunk_end)
```
acezen, line 672: Can we unify the three constructors into one and use default parameter like (vertex_chunk_begin = 0, vertex_chunk_end=std::numeric_limits<int64_t>::max()), then we can do some checks to match different cases:
if vertex_chunk_end == std::numeric_limits<int64_t>::max(): means the end is the last edge + 1 of last vertex chunk;

acezen: In the original constructors, there are not checking that begin and end are valid. maybe we need to add some check to ensure:
- begin <= vertex_chunk_end
- vertex_chunk_begin <= vertex_chunk_num && vertex_chunk_end <= vertex_chunk_num

lixueclaire: > Can we unify the three constructors into one and use default parameter like (vertex_chunk_begin = 0, vertex_chunk_end=std::numeric_limits<int64_t>::max()), then we can do some checks to match different cases: if vertex_chunk_end == std::numeric_limits<int64_t>::max(): means the end is the last edge + 1 of last vertex chunk;

Good suggestion! I have unified the constructors.

lixueclaire: > In the original constructors, there are not checking that begin and end are valid or not. maybe we need to add some check to ensure:
> 
> * begin <= vertex_chunk_end
> * vertex_chunk_begin <= vertex_chunk_num && vertex_chunk_end <= vertex_chunk_num

Perhaps we do not need to add checks here, for invalid input parameters for vertex_chunk_begin and vertex_chunk_end, an empty EdgesCollection will be returned. Related tests have been added.

acezen (approved):
LGTM~

Issue #140, opened 2023-04-18
[Feat]Complements the error message and handling of C++ and Spark


acezen:
TBF

acezen:
Make the ErrorCode specific

PR #192, opened 2023-06-25
[Improvement][C++] Refine the error message of errors of C++ SDK

acezen:
## Proposed changes
This change aims to refine the error constructor and fulfill the error message detail for C++ SDK.
- Refine the error constructor with [parameter pack](https://en.cppreference.com/w/cpp/language/parameter_pack), so developers can construct the error by only passing the messages to constructor, no need to concatenate as a string.
- Fulfill the detail message of error in C++ SDK to let user/developer know what the error is about and why it turn out.

## Types of changes

## Checklist

- [x] I have signed the CLA
- [x] I have added necessary documentation (if appropriate)

## Further comments
Related issue: #140

### DIFF
--- a/cpp/include/gar/graph_info.h
+++ b/cpp/include/gar/graph_info.h
@@ -583,8 +583,9 @@ class EdgeInfo {
   Status AddPropertyGroup(const PropertyGroup& property_group,
                           AdjListType adj_list_type) noexcept {
     if (!ContainAdjList(adj_list_type)) {
-      return Status::Invalid("Invalid adjacency list type: ",
-                             AdjListTypeToString(adj_list_type));
+      return Status::KeyError(
+          "Adjacency list type: ", AdjListTypeToString(adj_list_type),
+          " is not found in ", edge_label_, " edgeinfo.");
     }
     if (ContainPropertyGroup(property_group, adj_list_type)) {
       return Status::KeyError("Property group: ", property_group,
@@ -882,7 +883,7 @@ class EdgeInfo {
       IdType vertex_chunk_index, IdType edge_chunk_index) const {
     if (!ContainPropertyGroup(property_group, adj_list_type)) {
       return Status::KeyError("The property group: ", property_group,
-                              " is not found in the edge info");
+                              " is not found in the edge info.");
     }
     return prefix_ + adj_list2prefix_.at(adj_list_type) +
            property_group.GetPrefix() + "part" +
--- a/cpp/include/gar/graph_info.h
+++ b/cpp/include/gar/graph_info.h
@@ -133,6 +133,17 @@ class PropertyGroup {
    */
   inline const std::string& GetPrefix() const { return prefix_; }
 
+  friend std::ostream& operator<<(std::ostream& stream,
+                                  const PropertyGroup& pg) {
+    for (size_t i = 0; i < pg.properties_.size(); ++i) {
+      stream << pg.properties_[i].name;
+      if (i != pg.properties_.size() - 1) {
+        stream << "_";
+      }
+    }
+    return stream;
+  }
+
  private:
   std::vector<Property> properties_;
   FileType file_type_;
@@ -209,22 +220,22 @@ class VertexInfo {
    */
   inline Status AddPropertyGroup(const PropertyGroup& property_group) {
     if (ContainPropertyGroup(property_group)) {
-      return Status::InvalidOperation(
-          "The property group has already existed, can't not be added again.");
+      return Status::Invalid("The property group ", property_group,
+                             " has already existed, can't not be added again.");
     }
     for (const auto& property : property_group.GetProperties()) {
       if (ContainProperty(property.name)) {
-        std::string err_msg = "The property " + property.name +
-                              " has already existed in the vertex info.";
-        return Status::InvalidOperation(err_msg);
+        return Status::Invalid("The property ", property.name,
+                               " has already existed in the ", label_,
+                               " vertex info, can't not be added again.");
       }
     }
 
     property_groups_.push_back(property_group);
     for (const auto& p : property_group.GetProperties()) {
       if (!version_.CheckType(p.type.ToTypeName())) {
-        return Status::Invalid(
-            "The property type is not supported by the version.");
+        return Status::Invalid("The property type ", p.type.ToTypeName(),
+                               " is not supported by the version.");
       }
       p2type_[p.name] = p.type;
       p2primary_[p.name] = p.is_primary;
@@ -280,7 +291,8 @@ class VertexInfo {
   Result<const PropertyGroup&> GetPropertyGroup(
       const std::string& property_name) const noexcept {
     if (!ContainProperty(property_name)) {
-      return Status::KeyError("The property is not found.");
+      return Status::KeyError("No property with name ", property_name,
+                              " found in ", label_, " vertex.");
     }
     return property_groups_[p2group_index_.at(property_name)];
   }
@@ -295,7 +307,8 @@ class VertexInfo {
   inline Result<DataType> GetPropertyType(
       const std::string& property_name) const noexcept {
     if (p2type_.find(property_name) == p2type_.end()) {
-      return Status::KeyError("The property is not found.");
+      return Status::KeyError("No property with name ", property_name,
+                              " found in ", label_, " vertex.");
     }
     return p2type_.at(property_name);
   }
@@ -336,7 +349,8 @@ class VertexInfo {
   inline Result<bool> IsPrimaryKey(const std::string& property_name) const
       noexcept {
     if (p2primary_.find(property_name) == p2primary_.end()) {
-      return Status::KeyError("The property is not found.");
+      return Status::KeyError("No property with name ", property_name,
+                              " found in ", label_, " vertex.");
     }
     return p2primary_.at(property_name);
   }
@@ -383,8 +397,8 @@ class VertexInfo {
   inline Result<std::string> GetFilePath(const PropertyGroup& property_group,
                                          IdType chunk_index) const noexcept {
     if (!ContainPropertyGroup(property_group)) {
-      return Status::KeyError(
-          "Vertex info does not contain the property group.");
+      return Status::KeyError("The property group: ", property_group,
+                              " is not found in the ", label_, " vertex.");
     }
     return prefix_ + property_group.GetPrefix() + "chunk" +
            std::to_string(chunk_index);
@@ -400,8 +414,8 @@ class VertexInfo {
   inline Result<std::string> GetPathPrefix(
       const PropertyGroup& property_group) const noexcept {
     if (!ContainPropertyGroup(property_group)) {
-      return Status::KeyError(
-          "Vertex info does not contain the property group.");
+      return Status::KeyError("The property group: ", property_group,
+                              " is not found in the ", label_, " vertex.");
     }
     return prefix_ + property_group.GetPrefix();
   }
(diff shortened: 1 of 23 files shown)


acezen:
@lixueclaire The error messages maybe not very correct for writer and builder , or collections(just base on my understanding), can you take a look to the fulfill and if it's not correct, any improvement is good, thanks a lot.

lixueclaire:
> @lixueclaire The error messages maybe not very correct for writer and builder , or collections(just base on my understanding), can you take a look to the fulfill and if it's not correct, any improvement is good, thanks a lot.

Sure, I'll have a look soon. Thank you for making this change.

cpp/include/gar/utils/status.h
```diff
@@ -69,9 +91,6 @@ enum class StatusCode : unsigned char {
   kKeyError,
   kTypeError,
   kInvalid,
-  kInvalidValue,
```
lixueclaire, line 72: Could you please add some comments for the enum class to briefly introduce the types of different status code?

cpp/include/gar/graph_info.h
```diff
@@ -851,8 +881,8 @@ class EdgeInfo {
       const PropertyGroup& property_group, AdjListType adj_list_type,
       IdType vertex_chunk_index, IdType edge_chunk_index) const {
     if (!ContainPropertyGroup(property_group, adj_list_type)) {
-      return Status::KeyError(
-          "The edge info does not contain the property group.");
+      return Status::KeyError("The property group: ", property_group,
```
lixueclaire, line 884: Add "." in the end. The same for the following changes.

cpp/include/gar/reader/arrow_chunk_reader.h
```diff
@@ -84,7 +84,9 @@ class VertexPropertyArrowChunkReader {
       chunk_table_.reset();
     }
     if (chunk_index_ >= chunk_num_) {
-      return Status::KeyError();
```
lixueclaire, line 87: For me, it is a little confused to decide when to use `KeyError`, and when to use `Invalid`

cpp/include/gar/reader/arrow_chunk_reader.h
```diff
@@ -348,7 +359,10 @@ class AdjListOffsetArrowChunkReader {
       chunk_table_.reset();
     }
     if (chunk_index_ >= vertex_chunk_num_) {
-      return Status::KeyError();
+      return Status::KeyError("Internal vertex id ", id,
```
lixueclaire, line 362: I am not sure that why not to use  `OutofRange` here

cpp/src/graph_info.cc
```diff
@@ -108,7 +108,7 @@ Status VertexInfo::Save(const std::string& path) const {
 
 Result<EdgeInfo> EdgeInfo::Load(std::shared_ptr<Yaml> yaml) {
   if (yaml == nullptr) {
-    return Status::YamlError("yaml is nullptr");
+    return Status::Invalid("yaml is nullptr");
```
lixueclaire, line 111: modify the message to "yaml shared pointer is nullptr" to be consistent

cpp/src/chunk_info_reader.cc
```diff
@@ -52,13 +55,16 @@ Status AdjListChunkInfoReader::seek_src(IdType id) noexcept {
...
-        "The seek_dst operation is invalid in reader.");
+    return Status::Invalid("The seek_dst operation is invalid in edge ",
+                           edge_info_.GetEdgeLabel(), " reader with ",
+                           AdjListTypeToString(adj_list_type_), " type.");
   }
 
   IdType new_vertex_chunk_index = id / edge_info_.GetDstChunkSize();
   if (new_vertex_chunk_index >= vertex_chunk_num_) {
-    return Status::KeyError("The id " + std::to_string(id) + " not exist.");
+    return Status::KeyError("No vertex with internal id ", id,
```
lixueclaire, line 65: Also, this is not clear to me why we use `KeyError` here.

cpp/src/arrow_chunk_writer.cc
```diff
@@ -315,33 +318,37 @@ Status EdgeChunkWriter::validate(
...
   }
   // strong validate for the input_table
   if (validate_level == ValidateLevel::strong_validate) {
     auto schema = input_table->schema();
     int index = schema->GetFieldIndex(GeneralParams::kOffsetCol);
-    if (index == -1)
-      return Status::InvalidOperation("the offset column is not provided");
+    if (index == -1) {
+      return Status::Invalid("The offset column ", GeneralParams::kOffsetCol,
+                             " is not exist in the input table");
```
lixueclaire, line 346: does not exist

cpp/src/arrow_chunk_writer.cc
```diff
@@ -362,29 +369,36 @@ Status EdgeChunkWriter::validate(
...
+    if (index == -1) {
+      return Status::Invalid("The destination index column ",
+                             GeneralParams::kDstIndexCol,
+                             " is not exist in the input table");
+    }
     field = schema->field(index);
     if (field->type()->id() != arrow::Type::INT64) {
       return Status::TypeError(
-          "the data type for destination column should be INT64, but got " +
+          "The data type for destination column should be INT64, but got ",
```
lixueclaire, line 401: add "index" before "column" to be consistent

cpp/src/arrow_chunk_writer.cc
```diff
@@ -496,8 +515,10 @@ Status EdgeChunkWriter::WriteOffsetChunk(
   GAR_ASSIGN_OR_RAISE(auto file_type, edge_info_.GetFileType(adj_list_type_));
   auto schema = input_table->schema();
   int index = schema->GetFieldIndex(GeneralParams::kOffsetCol);
-  if (index == -1)
-    return Status::InvalidOperation("the offset column is not provided");
+  if (index == -1) {
+    return Status::Invalid("The offset column ", GeneralParams::kOffsetCol,
+                           " is not exist in the input table");
```
lixueclaire, line 520: does not exist

cpp/src/arrow_chunk_writer.cc
```diff
@@ -515,12 +536,18 @@ Status EdgeChunkWriter::WriteAdjListChunk(
   indices.clear();
   auto schema = input_table->schema();
   int index = schema->GetFieldIndex(GeneralParams::kSrcIndexCol);
-  if (index == -1)
-    return Status::InvalidOperation("the source column is not provided");
+  if (index == -1) {
+    return Status::Invalid("The source index column ",
+                           GeneralParams::kSrcIndexCol,
+                           " is not exist in the input table");
```
lixueclaire, line 542: ditto

cpp/src/arrow_chunk_writer.cc
```diff
@@ -515,12 +536,18 @@ Status EdgeChunkWriter::WriteAdjListChunk(
...
+  }
   indices.push_back(index);
   index = schema->GetFieldIndex(GeneralParams::kDstIndexCol);
-  if (index == -1)
-    return Status::InvalidOperation("the destination column is not provided");
+  if (index == -1) {
+    return Status::Invalid("The destination index column ",
+                           GeneralParams::kDstIndexCol,
+                           " is not exist in the input table");
```
lixueclaire, line 549: ditto

cpp/src/arrow_chunk_writer.cc
```diff
@@ -545,9 +572,10 @@ Status EdgeChunkWriter::WritePropertyChunk(
   for (auto& property : property_group.GetProperties()) {
     int indice = schema->GetFieldIndex(property.name);
     if (indice == -1) {
-      std::string msg = "invalid property: " + property.name +
-                        " edge: " + edge_info_.GetEdgeLabel();
-      return Status::InvalidOperation(msg);
+      return Status::Invalid("Property ", property.name, " of property group ",
+                             property_group, " of edge ",
+                             edge_info_.GetEdgeLabel(),
+                             " is not exist in the input table.");
```
lixueclaire, line 578: ditto

cpp/include/gar/reader/arrow_chunk_reader.h
acezen: I think, `KeyError` should use in the scenario that user give a key, we can't find a key in our structure like that, otherwise, use a `Invalid`, so here raise `Invalid` is better. Thanks for question, I will fix it

cpp/include/gar/reader/arrow_chunk_reader.h
acezen: good point, as above mention, I think I confused the KeyError and OutOfRange here, in the `seek` scenario, the given parameter is in a range and it should raise a OutOfRange here. I will double check all the `seek` operation.

cpp/include/gar/reader/arrow_chunk_reader.h
lixueclaire: > I think, `KeyError` should use in the scenario that user give a key, we can't find a key in our structure like that, otherwise, use a `Invalid`, so here raise `Invalid` is better. Thanks for question, I will fix it

Then, it is not very clear that which can be called as a "Key": internal vertex id? chunk index? property name? property group? vertex/edge label? adj_list type? ....

cpp/include/gar/utils/status.h
lixueclaire: like this: https://arrow.apache.org/docs/cpp/api/support.html#error-return-and-reporting

cpp/src/arrow_chunk_writer.cc
```diff
@@ -362,29 +369,37 @@ Status EdgeChunkWriter::validate(
...
+  // strong validate for the input table
   if (validate_level == ValidateLevel::strong_validate) {
     auto schema = input_table->schema();
     int index = schema->GetFieldIndex(GeneralParams::kSrcIndexCol);
-    if (index == -1)
-      return Status::InvalidOperation("the source column is not provided");
+    if (index == -1) {
+      return Status::Invalid("The source index column ",
+                             GeneralParams::kSrcIndexCol,
+                             " is not exist in the input table");
```
lixueclaire, line 384: does not

cpp/src/arrow_chunk_writer.cc
```diff
@@ -362,29 +369,37 @@ Status EdgeChunkWriter::validate(
...
+          "The data type for source index column should be INT64, but got ",
           field->type()->name());
     }
     index = schema->GetFieldIndex(GeneralParams::kDstIndexCol);
-    if (index == -1)
-      return Status::InvalidOperation("the destination column is not provided");
+    if (index == -1) {
+      return Status::Invalid("The destination index column ",
+                             GeneralParams::kDstIndexCol,
+                             " is not exist in the input table");
```
lixueclaire, line 396: does not

cpp/src/arrow_chunk_writer.cc
```diff
@@ -121,49 +117,54 @@ Status VertexPropertyWriter::validate(const PropertyGroup& property_group,
...
   // weak & strong validate
   if (!vertex_info_.ContainPropertyGroup(property_group)) {
-    return Status::InvalidOperation(
-        "the property group does not exist in the vertex info");
+    return Status::KeyError("The property group ", property_group,
+                            " does not exist in ", vertex_info_.GetLabel(),
+                            " vertex info.");
+  }
+  if (chunk_index < 0) {
+    return Status::Invalid("The chunk index is negative.");
```
lixueclaire, line 125: I think IndexError is more suitable here.

lixueclaire (approved):

cpp/src/arrow_chunk_writer.cc
lixueclaire: And the unit test need to be updated too.

Issue #72, opened 2023-01-11
[Feat] Provide libraries for other languages


acezen:
**Is your feature request related to a problem? Please describe.**
Currently the libraries for GraphAr are only available for C++ and Spark. But many graph processing systems are implemented by other programming languages (like Neo4j by java). We need to provide libraries for more programming languages.

**Describe the solution you'd like**
Implement library with
- [ ] Java
- [ ] Go
- [ ] Rust
- [ ] Python

kesavkolla:
Any movement on this? Maybe support a webassembly binding so that we can use it from any other programming language. WASI is now a days is the default binding.

acezen:
> Any movement on this? Maybe support a webassembly binding so that we can use it from any other programming language. WASI is now a days is the default binding.

Hi, @kesavkolla , thanks you for your interest in GraphAr. The work of implement libraries for languages  is not start yet but it would be our next major work of development.

I'm not very familiar with webassembly binding but according to the document it seems to be a great solution to make GraphAr accessible for use with programming languages other than C++ and Spark. It would be very helpful if you can provide some detail idea about webassembly binding and how do you want to use GraphAr?

kesavkolla:
@acezen I am seeing this as a trend with many open source projects where the base SDK is developed in one language either C or rust and other language SDKs are derived from WASI. WASI has good compatible layer with many implementations like wasmer, wasmtime etc... it will be so easy to use in any programming language. Investing in each programming language as a separate SDK requires lot of effort in each programming language. So IMO wasm is the right choice to go with.

kesavkolla:
I am actually looking for Rust client so that I can load GraphAr in rust and start leveraging Gremlin path traversal.

kesavkolla:
@acezen I have asked a question in hugegraph community about support for GraphAr
https://github.com/apache/incubator-hugegraph/discussions/2149

They also say GraphAR to support WASM binding. IMO This is really a huge thing so that many programming languages can start using GraphAR.

sighingnow:
Hi @kesavkolla,

> I am actually looking for Rust client so that I can load GraphAr in rust and start leveraging Gremlin path traversal.

Yes, we also have the plan to add rust support to read and build graph data in GraphAr format. We currently have a on-going refactor (#116) to have a clear repo layout for multi-languages supported. We could launch the Rust SDK after that PR merged and it would nice if we could have some collaboration to develop the Rust SDK.

[...]

PR #233, opened 2023-09-17
[FEAT][JAVA] Implement READERS and WRITERS for Java

Thespica:
## Proposed changes

As titile.

Next step: Add related documents and library referance.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

Related issue: #72


.github/workflows/java.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           
       - name: Code Format Check
         run: |
-          export JAVA_HOME=${JAVA_HOME_11_X64}
+          export JAVA_HOME=${JAVA_HOME_8_X64}
```
acezen, line 46: why change to JAVA 8?

java/CMakeLists.txt
```diff
@@ -16,10 +18,12 @@ include_directories(SYSTEM ${JAVA_INCLUDE_PATH2})
 include_directories("src/main/native")
 include_directories("src/test/native")
 
+find_package(Arrow REQUIRED)
 find_package(gar REQUIRED)
 
 add_library(${LIBNAME} SHARED ${SOURCES})
 target_link_libraries(${LIBNAME} ${CMAKE_JNI_LINKER_FLAGS} gar)
+target_link_libraries(${LIBNAME} ${CMAKE_JNI_LINKER_FLAGS} Arrow::arrow_static)
```
acezen, line 26: maybe link to shared library is better? otherwise it would take a long time to compile

java/pom.xml
```diff
@@ -16,6 +16,7 @@
         <maven.compiler.target>8</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <fastffi.revision>0.1.2</fastffi.revision>
+        <arrow.version>13.0.0</arrow.version>
```
acezen, line 19: Does this have to be set? If user install other version of arrow,  the program still work as expect?

java/src/main/java/com/alibaba/graphar/readers/arrowchunk/VertexPropertyArrowChunkReader.java
```diff
@@ -0,0 +1,130 @@
...
+                @CXXReference StdString prefix);
+
+        /**
+         * Initialize the VertexPropertyArrowChunkReader.
+         *
+         * @param vertexInfo The vertex info that describes the vertex type.
+         * @param propertyGroup The property group that describes the property group.
+         * @param prefix The absolute prefix.
+         */
+        //    VertexPropertyArrowChunkReader create(
```
acezen, line 123: Can you add some comment here if the code has to remain?

.github/workflows/java.yml
Thespica: I haven't used features that higher than Java 8, and Java 8 have better compatibility with others graph systems which rely on Java 8. So maybe we should assume that environment is Java 8, and Java higher is compatible with lower version.

java/CMakeLists.txt
Thespica: On one head, linking arrow as dynamic will leads error about libprotobuf: https://github.com/Thespica/GraphAr/actions/runs/6206365474/job/16850543098#step:6:9100

On the other head, I consider linking arrow as static just copy arrow lib once, rather than recompile, which only cost little time.

java/src/main/java/com/alibaba/graphar/readers/arrowchunk/VertexPropertyArrowChunkReader.java
Thespica: Removed

java/pom.xml
acezen: if work as expected, 13.0.0 is ok

.github/workflows/java.yml
acezen: the spark rely on java 11, so for CI unification and avoid confusion， I think 11 is better.

java/CMakeLists.txt
acezen: OK, this can remain and open as a issue to fix in the future.

acezen (approved):
Good job! There're still some tiny problem to fix(see comments), and when it done, the PR if fine to me to merge to main

PR #254, opened 2023-10-18
[Minor][Doc] Add release-process.md to explain the release process

acezen:
## Proposed changes
Since road map is a high level vision of GraphAr, we need a document to explain the release process and milestone detail, as a supplement of road map。


lixueclaire (approved):
LGTM

Issue #287, opened 2023-11-24
[Feat][Format] Support List Data Type


acezen:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #169, opened 2023-05-24
[Feat] Support `Date` and `List` standard data type for GraphAr


acezen:
**Is your feature request related to a problem? Please describe.**
As a graph format , GraphAr supports a variety of data types, including strings, numbers, and booleans. However, it does not currently support the Date and List standard data types. This issue proposes adding support for these data types to GraphAr.

Adding support for Date and Array would make GraphAr a more versatile. It would allow users to store and query dates and arrays, which are common data types in many graph database and system.

**Describe the solution you'd like**
The `Date` could use `int32` or `int64` to represent.  Just unify it in `parquet` and `orc`;
And the `List`  could be use `parquet`'s BYTE_ARRAY and `ORC`'s list;

The `serialization` and `deserialization` strategy should be define for the `Data` and `List` type.

**Additional context**
This issue is a part of issue #76  and is a good first issue for beginners to get familiar with GraphAr.

varuntumbe:
Hi @acezen , Can I try my hands on this issue ? I might get lot of questions, will ask in discord group

Thanks

lixueclaire:
> Hi @acezen , Can I try my hands on this issue ? I might get lot of questions, will ask in discord group
> 
> Thanks

Thanks for your interest in contributing to our project! We'd be happy to have you work on this issue.
If you have any questions or need help getting started, please don't hesitate to ask. Feel free to ask questions through issues, discussions, e-mail, or Slack. Welcome to the [GraphAr Community](https://github.com/alibaba/GraphAr/blob/main/docs/developers/community.rst), and we're looking forward to your contributions.

acezen:
> 

Sure!,  any contribution is welcome. As @lixueclaire said, if you have any question or need help, fell free to ask~

varuntumbe:
Thanks ! I will take like a week to kinda understand the whole thing and setup it in local, once done will ping here for more info.

Thanks and Regards 
Varun

acezen:
The `List` data type has been support with Pull Requet #296

PR #296, opened 2023-12-19
[Feat][C++] Support List Data Type, use `list<float>` as example

acezen:
## Proposed changes
This pull request aims to support `List` data type in GraphAr C++ and add related unit-test
- Support `List` data type by add a `child_` to `DataType` class, implement by nested other logical data type.
- In high level reader, support read list property  in `Vertex`/`Edge`.
yaml example:
```bash
property:
  - name: feature
  - data_type: list<float>
  - is_primary: false
```

NB: since BooleanArray and StringArray'method are not align to NumericalArray, so we only support `list<int32>` `list<int64>`, `list<float>` and `list<double>` at present.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments
close issue #287 
related to #169

### DIFF
--- a/cpp/include/gar/graph.h
+++ b/cpp/include/gar/graph.h
@@ -70,23 +70,7 @@ class Vertex {
    * @return Result: The property value or error.
    */
   template <typename T>
-  inline Result<T> property(const std::string& property) noexcept {
-    if (properties_.find(property) == properties_.end()) {
-      return Status::KeyError("Property with name ", property,
-                              " does not exist in the vertex.");
-    }
-    try {
-      T ret = std::any_cast<T>(properties_[property]);
-      return ret;
-    } catch (const std::bad_any_cast& e) {
-      return Status::TypeError("Any cast failed, the property type of ",
-                               property, " is not matched ", e.what());
-    }
-  }
-
-  template <typename T>
-  Result<std::pair<const T*, int>> list_property(
-      const std::string& property) const;
+  Result<T> property(const std::string& property) const;
 
  private:
   IdType id_;
@@ -130,23 +114,7 @@ class Edge {
    * @return Result: The property value or error.
    */
   template <typename T>
-  inline Result<T> property(const std::string& property) noexcept {
-    if (properties_.find(property) == properties_.end()) {
-      return Status::KeyError("Property with name ", property,
-                              " does not exist in the edge.");
-    }
-    try {
-      T ret = std::any_cast<T>(properties_[property]);
-      return ret;
-    } catch (const std::bad_any_cast& e) {
-      return Status::TypeError("Any cast failed, the property type of ",
-                               property, " is not matched ", e.what());
-    }
-  }
-
-  template <typename T>
-  Result<std::pair<const T*, int>> list_property(
-      const std::string& property) const;
+  inline Result<T> property(const std::string& property) const;
 
  private:
   IdType src_id_, dst_id_;
--- a/cpp/include/gar/graph.h
+++ b/cpp/include/gar/graph.h
@@ -85,7 +85,8 @@ class Vertex {
   }
 
   template <typename T>
-  Result<std::pair<const T*, int>> list_property(const std::string& property);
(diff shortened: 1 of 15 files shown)


cpp/include/gar/graph.h
```diff
@@ -83,9 +84,14 @@ class Vertex {
     }
   }
 
+  template <typename T>
+  Result<std::pair<const T*, int>> list_property(
```
acezen, line 88: Maybe we should add a `List` class and unify the `list_property` and `property`

acezen: update

acezen:
hi，@lixueclaire , Do you want to give this a review?

lixueclaire:
> hi，@lixueclaire , Do you want to give this a review?

OK, I'll have a look right now.

cpp/include/gar/util/data_type.h
```diff
@@ -51,6 +51,9 @@ enum class Type {
   /** UTF8 variable-length string */
   STRING,
 
+  /** List of some logical data type */
+  LIST,
```
lixueclaire, line 55: hi, @acezen . In C++, the logical data type is LIST; and in Spark, which we implemented previously, it is called ARRAY (https://github.com/alibaba/GraphAr/blob/main/spark/src/main/scala/com/alibaba/graphar/GraphInfo.scala#L65), should we unify this?

cpp/test/test_graph.cc
```diff
@@ -80,6 +80,31 @@ TEST_CASE("Graph") {
...
+    auto graph_info = maybe_graph_info.value();
+    std::string label = "person", list_property = "feature";
+    auto maybe_vertices_collection =
+        VerticesCollection::Make(graph_info, label);
+    REQUIRE(!maybe_vertices_collection.has_error());
+    auto vertices = maybe_vertices_collection.value();
+    auto count = 0;
+    for (auto it = vertices->begin(); it != vertices->end(); ++it) {
+      auto vertex = *it;
+      auto float_array = vertex.property<FloatArray>(list_property).value();
```
lixueclaire, line 98: I think it could be better to acquire the datatype of list_property first before getting the value.

cpp/include/gar/util/data_type.h
```diff
@@ -51,6 +51,9 @@ enum class Type {
   /** UTF8 variable-length string */
   STRING,
 
+  /** List of some logical data type */
```
lixueclaire, line 54: Could you please add LIST into the [data types](https://alibaba.github.io/GraphAr/file-format.html#data-types) and explain that CSV does not support it?

cpp/src/vertices_builder.cc
```diff
@@ -65,37 +65,37 @@ Status VerticesBuilder::validate(const Vertex& v, IdType index,
       switch (type->id()) {
       case Type::BOOL:
         if (property.second.type() !=
-            typeid(typename ConvertToArrowType<Type::BOOL>::CType)) {
+            typeid(typename TypeToArrowType<Type::BOOL>::CType)) {
```
lixueclaire, line 68: I want to understand that if we should add Type::LIST into the implementations of VerticesBuilder/EdgesBuilder.

cpp/include/gar/util/data_type.h
acezen: fixed

cpp/test/test_graph.cc
acezen: done

cpp/src/vertices_builder.cc
acezen: Good question, I think this can add to Builder but it's a little complicated at present. I can add a TODO tag and create a new issue.

lixueclaire: > Good question, I think this can add to Builder but it's a little complicated at present. I can add a TODO tag and create a new issue.

I agree, and I have no further questions.

lixueclaire (approved):

PR #314, opened 2024-01-03
Change VertexWriter constructor signature

### LINKED ISSUE #313: [Feat][Spark] Replace scala.Option[Long] in constructor of VertexWriter due to incompatibility with py4j
**Is your feature request related to a problem? Please describe.**
Currently the `scala.Option[Long]` [is used](https://github.com/alibaba/GraphAr/blob/main/spark/src/main/scala/com/alibaba/graphar/writer/VertexWriter.scala#L71) in constructor of `VertexWriter`. It is good from the scala point of view but there is a problem when you try to call this constructor from python with `py4j`. The reason is that `py4j` make [autoboxing and autounboxing](https://www.py4j.org/advanced_topics.html#boxing) of Java primitives. Let's imagine we have `py4j.java_gateway.JVMView` object in python, that is name `jvm`. Even if you try to pass into constructor something like `jvm.scala.Some(jvm.java.lang.Long.valueOf(100))` it will run the instruction in parentheses first, get `java.lang.Long` as a result, make autounboxing it into python `int` and pass `int` into `Some` with autoboxing it to `java.lang.Integer` that produce `Some[Int]`, not `Some[Long]`. And, because of using `Option` in constructor, `py4j` cannot understand what is expected type here and cannot make autoboxing of `int` into `jvm.java.lang.Long`. I tried different tricks, but each time I'm getting `java.lang.Integer cannot be cast to java.lang.Long`. There is [a similar discussion](https://github.com/py4j/py4j/issues/374#issuecomment-513463892) in `py4j` issues, which is open from 2019 and it looks like there is no solution. So, currently, the only option to make a `VertexWriter` from python is pass `jvm.scala.Option.empty()` and no way to pass optional parameter `numVertices`.

**Describe the solution you'd like**
Make tow constructor of `VertexWriter`. The first one like `VertexWriter(prefix: String, vertexInfo: VertexInfo, vertexDf: DataFrame, numVertices: Long)` and the second one like `VertexWriter(prefix: String, vertexInfo: VertexInfo, vertexDf: DataFrame)`.

**Describe alternatives you've considered**
Alternative solution is to add an additional method into `VertexInfo` companion object, like `createFromPy4j(prefix: String, vertexInfo: VertexInfo, vertexDf: DataFrame, numVertices: Long): VertexInfo`. It allows to decide on the python side, which one to call: if python user provide `numVertices` than call a fabric-method from companion object, otherwise call constructor with `jvm.scala.Option.empty()`.

**Additional context**
I can do it by myself. I'm sure, that no additional changes are required and all spark scala API will be the same.

P.S. I understand, that such a way is not a "scala way", but in my experience each spark library sooner or later face the question of creating PySpark bindings. It looks OK for me to sacrifice some scala code beauty to get better compatibility with PySpark...

SemyonSinchenko:
## Proposed changes

Changes in constructor of `VertexWriter` as we discussed in #313 

## Checklist

- [x] I have signed the CLA
- [x] I have added necessary documentation (if appropriate)

## Further comments

Close #313 
Related to #300

### DIFF
--- a/spark/src/main/scala/com/alibaba/graphar/graph/GraphWriter.scala
+++ b/spark/src/main/scala/com/alibaba/graphar/graph/GraphWriter.scala
@@ -94,7 +94,7 @@ class GraphWriter() {
         val df_with_index = df_and_mapping._1
         indexMappings += label -> df_and_mapping._2
         val writer =
-          new VertexWriter(prefix, vertexInfo, df_with_index, Some(vertex_num))
+          new VertexWriter(prefix, vertexInfo, df_with_index, vertex_num)
         writer.writeVertexProperties()
       }
     }
--- a/spark/src/main/scala/com/alibaba/graphar/writer/VertexWriter.scala
+++ b/spark/src/main/scala/com/alibaba/graphar/writer/VertexWriter.scala
@@ -63,19 +63,20 @@ object VertexWriter {
  *   the vertex info that describes the vertex type.
  * @param vertexDf
  *   the input vertex DataFrame.
+ * @param numVertices
+ *   the number of vertices, if negative value is passed, then vertexDF.count
+ *   will be used
  */
 class VertexWriter(
     prefix: String,
     vertexInfo: VertexInfo,
     vertexDf: DataFrame,
-    numVertices: Option[Long] = None
+    numVertices: Long = -1
 ) {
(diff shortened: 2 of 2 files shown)


acezen (approved):
LGTM， thanks for the work!

Issue #347, opened 2024-01-26
[Bug][Spark]: Neo4j2GraphAr change the type


jasinliu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

<img width="875" alt="d8cf11b60af5db47e4a6d9f7af4955c8" src="https://github.com/alibaba/GraphAr/assets/85552719/2c54a126-df9b-465e-828d-d227e8a542a4">

The former data type is Long in neo4j. But after importer, it becomes string.

Just like [this](https://github.com/GraphScope/gar-test/blob/main/neo4j/Person.vertex.yml#L11).

### Expected Behavior

Fix it.

### Minimal Reproducible Example

Maybe [this](https://github.com/GraphScope/gar-test/blob/main/neo4j/Person.vertex.yml#L11).

### Environment

- Operating system: ubuntu
- GraphAr version: latest

jasinliu:
![image](https://github.com/alibaba/GraphAr/assets/85552719/94d112fd-8e11-4915-91c0-3c966d1aecd7)
When do import, the first warnning log maybe a key infomation as the image shows.

acezen:
Maybe we can use the [APOC procedure](https://neo4j.com/docs/spark/current/architecture/#_apoc), but it only available in 
enterprise edition

acezen:
And anther solution is change the option [schema.flatten.limit](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) to 1(default is 10). I change the code
https://github.com/alibaba/GraphAr/blob/fe4ebb9b2dbcf30e63cd3895a51d4d614fe80df3/spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala#L75-L78
to
```java
    val person_df = spark.read
      .format("org.neo4j.spark.DataSource")
      .option("schema.flatten.limit", 1)
      .option("query", "MATCH (n:Person) RETURN n.name AS name, n.born as born")
      .load()
```
and it work, the schema of person_df turns to be
```
root
 |-- name: string (nullable = true)
 |-- born: long (nullable = true)
```

acezen:
And I think this is not the bug of GraphAr, so I change the title and label to `improvement`.

PR #353, opened 2024-02-01
[Improvement][Spark] Try to make neo4j generate DataFrame with the correct data type

acezen:
## Proposed changes

related issue #347 
## Checklist

- [ ] I have signed the CLA

## Further comments

### DIFF
--- a/spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
+++ b/spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
@@ -72,9 +72,12 @@ object Neo4j2GraphAr {
       spark: SparkSession
   ): Unit = {
     // read vertices with label "Person" from Neo4j as a DataFrame
+    // Note: set "schema.flatten.limit" to 1 to avoid select null to make the property type as string,
+    // If APOC is installed, you can use apoc to get the property type
     val person_df = spark.read
       .format("org.neo4j.spark.DataSource")
       .option("query", "MATCH (n:Person) RETURN n.name AS name, n.born as born")
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, vertex label is "Person"
     writer.PutVertexData("Person", person_df)
@@ -86,6 +89,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (n:Movie) RETURN n.title AS title, n.tagline as tagline"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, vertex label is "Movie"
     writer.PutVertexData("Movie", movie_df)
@@ -97,6 +101,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:PRODUCED]->(b:Movie) return a.name as src, b.title as dst"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "PRODUCED"
     // target vertex label is "Movie"
@@ -109,6 +114,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:ACTED_IN]->(b:Movie) return a.name as src, b.title as dst"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "ACTED_IN"
     // target vertex label is "Movie"
@@ -121,6 +127,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:DIRECTED]->(b:Movie) return a.name as src, b.title as dst"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "DIRECTED"
     // target vertex label is "Movie"
@@ -133,6 +140,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:FOLLOWS]->(b:Person) return a.name as src, b.name as dst"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "FOLLOWS"
     // target vertex label is "Person"
@@ -145,6 +153,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:REVIEWED]->(b:Movie) return a.name as src, b.title as dst, r.rating as rating, r.summary as summary"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "REVIEWED"
     // target vertex label is "Movie"
@@ -157,6 +166,7 @@ object Neo4j2GraphAr {
         "query",
         "MATCH (a:Person)-[r:WROTE]->(b:Movie) return a.name as src, b.title as dst"
       )
+      .option("schema.flatten.limit", 1)
       .load()
     // put into writer, source vertex label is "Person", edge label is "WROTE"
     // target vertex label is "Movie"
--- a/spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
+++ b/spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
@@ -72,8 +72,8 @@ object Neo4j2GraphAr {
       spark: SparkSession
   ): Unit = {
     // read vertices with label "Person" from Neo4j as a DataFrame
-    // Note: set "schema.flatten.limit" to 1 to avoid select null to make the property type as string,
-    // If APOC is installed, you can use apoc to get the property type
+    // Note: set "schema.flatten.limit" to 1 to not sample null record infer type as string as far as possible,
+    // if you want a perfect type inference, consider to user APOC.
     val person_df = spark.read
       .format("org.neo4j.spark.DataSource")
       .option("query", "MATCH (n:Person) RETURN n.name AS name, n.born as born")


spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
```diff
@@ -72,9 +72,12 @@ object Neo4j2GraphAr {
       spark: SparkSession
   ): Unit = {
     // read vertices with label "Person" from Neo4j as a DataFrame
+    // Note: set "schema.flatten.limit" to 1 to avoid select null to make the property type as string,
+    // If APOC is installed, you can use apoc to get the property type
     val person_df = spark.read
       .format("org.neo4j.spark.DataSource")
       .option("query", "MATCH (n:Person) RETURN n.name AS name, n.born as born")
+      .option("schema.flatten.limit", 1)
```
lixueclaire, line 80: I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?

acezen: > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?

we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism 
 of neo4j, this is the best current solution we can provide for example.(except use the APOC)

What case you suggest to add a test case?

lixueclaire: > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> 
> we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> 
> What case you suggest to add a test case?

My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.

acezen: > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > 
> > 
> > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > What case you suggest to add a test case?
> 
> My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.

The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct record is  in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.

lixueclaire: > > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > > 
> > > 
> > > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > > What case you suggest to add a test case?
> > 
> > 
> > My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.
> 
> The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct records are in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.

I understand Neo4j's current approach to type inference, and that correct records are in majority. Is it possible to use the first non-null value for type inference? This could enhance the robustness of the current schema.

acezen: > > > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > > > 
> > > > 
> > > > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > > > What case you suggest to add a test case?
> > > 
> > > 
> > > My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.
> > 
> > 
> > The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct records are in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.
> 
> I understand Neo4j's current approach to type inference, and that correct records are in majority. Is it possible to use the first non-null value for type inference? This could enhance the robustness of the current schema.

As I know, Neo4j does not provide such approach to infer data type except using APOC(but apoc feature is ononly available in enterprise edition).

lixueclaire: > > > > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > > > > 
> > > > > 
> > > > > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > > > > What case you suggest to add a test case?
> > > > 
> > > > 
> > > > My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.
> > > 
> > > 
> > > The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct records are in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.
> > 
> > 
> > I understand Neo4j's current approach to type inference, and that correct records are in majority. Is it possible to use the first non-null value for type inference? This could enhance the robustness of the current schema.
> 
> As I know, Neo4j does not provide such approach to infer data type except using APOC(but apoc feature is ononly available in enterprise edition).

I think a possible way is to apply the cast method (of spark) with DataType on the column of the DataFrame. If implementing this solution is deemed too complex for our current scope, we can record this issue for future reference and potential enhancement.

acezen: > > > > > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > > > > > 
> > > > > > 
> > > > > > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > > > > > What case you suggest to add a test case?
> > > > > 
> > > > > 
> > > > > My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.
> > > > 
> > > > 
> > > > The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct records are in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.
> > > 
> > > 
> > > I understand Neo4j's current approach to type inference, and that correct records are in majority. Is it possible to use the first non-null value for type inference? This could enhance the robustness of the current schema.
> > 
> > 
> > As I know, Neo4j does not provide such approach to infer data type except using APOC(but apoc feature is ononly available in enterprise edition).
> 
> I think a possible way is to apply the cast method (of spark) with DataType on the column of the DataFrame. If implementing this solution is deemed too complex for our current scope, we can record this issue for future reference and potential enhancement.

In my opinion, I think it's Neo4j's duty to ensure the schema's correct( and they put the feature to enterprise edition, is reasonable, to make user to pay money), we should not put these part to GraphAr, and we just need to make sure that the schema is consistent from DataFrame to the graphar files.

lixueclaire: > > > > > > > > I'm concerned that if the first value is null, the returned result might be incorrect. Could we add a test case to verify this case?
> > > > > > > 
> > > > > > > 
> > > > > > > we can not avoid such case since Neo4j use result to deduce the schema and we can not change the mechanism of neo4j, this is the best solution we can provide for example.(except use the APOC)
> > > > > > > What case you suggest to add a test case?
> > > > > > 
> > > > > > 
> > > > > > My concern is specifically with cases where the first value is null, but subsequent values are of type Long. Ideally, we should still infer the type as Long. However, I suspect that with schema.flatten.limit=1, the inferred type might default to String. I recommend we add a test to confirm the actual behavior in this scenario.
> > > > > 
> > > > > 
> > > > > The [schema.strategy](https://neo4j.com/docs/spark/current/reading/#_neo4j_read_options) would make neo4j sample `schema.flatten.limit` records to infer type. This change is just minish the `schema.flatten.limit`(default is 10) to let neo4j sample the right record as far as possible( we assume that correct records are in majority）. Sadly we can not guarantee that the test would always sample a Long record if we add.
> > > > 
> > > > 
> > > > I understand Neo4j's current approach to type inference, and that correct records are in majority. Is it possible to use the first non-null value for type inference? This could enhance the robustness of the current schema.
> > > 
> > > 
> > > As I know, Neo4j does not provide such approach to infer data type except using APOC(but apoc feature is ononly available in enterprise edition).
> > 
> > 
> > I think a possible way is to apply the cast method (of spark) with DataType on the column of the DataFrame. If implementing this solution is deemed too complex for our current scope, we can record this issue for future reference and potential enhancement.
> 
> In my opinion, it's Neo4j's duty to ensure the schema's correct( and they put the feature to enterprise edition, is reasonable, to make user to pay money), we should not put these part to GraphAr and it's not GraphAr's scope. GraphAr just need to make sure that the schema is consistent from DataFrame to the graphar files.

Understand you perspective. With respect to the schema.flatten.limit setting, although adjusting it might help in this  particular scenario, and I agree with modifying it in this example, I think it does not signify a bug within GraphAr.

lixueclaire (approved):

spark/src/main/scala/com/alibaba/graphar/example/Neo4j2GraphAr.scala
acezen: agree, I would retitle the PR to `improvement`.

PR #391, opened 2024-03-07
[Minor][C++] Revise the unsupported data type error msg to give more information

acezen:
## Proposed changes
as title

## Checklist

- [ ] I have signed the CLA

## Further comments

### DIFF
--- a/cpp/src/data_type.cc
+++ b/cpp/src/data_type.cc
@@ -44,6 +44,7 @@ std::shared_ptr<arrow::DataType> DataType::DataTypeToArrowDataType(
   default:
     std::string msg = "The data type " + type->ToTypeName() +
                       " is not supported yet in GraphAr";
+    throw std::runtime_error(msg);
   }
 }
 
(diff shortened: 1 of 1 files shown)


cpp/src/data_type.cc
```diff
@@ -42,7 +42,8 @@ std::shared_ptr<arrow::DataType> DataType::DataTypeToArrowDataType(
   case Type::LIST:
     return arrow::list(DataTypeToArrowDataType(type->child_));
   default:
-    throw std::runtime_error("Unsupported data type");
+    std::string msg = "The data type " + type->ToTypeName() +
```
Thespica, line 45: I wonder how the default case return or throw, because it doesn't `return` or `throw` directly.

lixueclaire: throw std::runtime_error(msg);

acezen: Thanks for pointing out, fixed.

lixueclaire (approved):
LGTM

PR #415, opened 2024-03-26
[FEAT][C++] Not allow setting custom namespace for code clarity

### LINKED ISSUE #414: [Feat][C++] Remove the  custom namespace setting feature
**Is your feature request related to a problem? Please describe.**
Currently GraphAr C++ provides support user to set custom namespace to avoid name conflict. But since the namespace `graphar` is not a widely-used names, I suggest we can remove the feature.

Allowing users to set a custom namespace for a C++ library during the CMake configuration step is not considered a best practice in the C++ community.Here are several reasons why this approach is generally discouraged:

- **Consistency**: A consistent namespace across all deployments of the library aids in maintainability and user understanding. Changing namespaces can lead to confusion, especially when dealing with community support or third-party tutorials and examples.
- **Documentation**: Documentation becomes more challenging to write and maintain if the namespace can vary, as examples in the documentation might not match the user's configured namespace.
- **Code Clarity**: When working in a codebase that uses the library, having a standard namespace makes it clear which library the code is interfacing with.
- **Tooling and Integration**: Tools that perform static analysis, code indexing, and refactoring may not be aware of the dynamic namespace configuration, leading to incorrect behavior.
-**Build Systems and Package Managers**: Package managers and build systems expect library names and namespaces to be consistent to manage dependencies correctly.
- **Namespace Collisions**: The primary motivation for allowing configurable namespaces might be to avoid namespace collisions. However, such collisions are rare, and when they occur, they can often be resolved using namespace aliases or other techniques that don't require changing the library's namespace.

We should make the design right in the GraphAr's beginning

**Describe the solution you'd like**
Remove the custom namespace setting feature.

**Describe alternatives you've considered**

**Additional context**

acezen:
## Proposed changes
As issue #414 describe

## Checklist

- [x] I have signed the CLA

## Further comments

close #414

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -37,16 +37,10 @@ project(graph-archive LANGUAGES C CXX VERSION ${GAR_VERSION})
 # cmake options
 # ------------------------------------------------------------------------------
 
-option(NAMESPACE "User specific namespace, default is graphar" OFF)
 option(BUILD_TESTS "Build unit tests" OFF)
 option(BUILD_EXAMPLES "Build examples" OFF)
 option(BUILD_BENCHMARKS "Build benchmarks" OFF)
 
-if (NAMESPACE)
-  add_definitions(-DGAR_NAMESPACE=${NAMESPACE})
-else()
-  add_definitions(-DGAR_NAMESPACE=graphar)
-endif()
 # ------------------------------------------------------------------------------
 # setting default cmake type to Release
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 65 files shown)


lixueclaire (commented):
Could you please update the examples in documentations to use `graphar` instead of `GAR_NAMESPACE`?

acezen:
> Could you please update the examples in documentations to use `graphar` instead of `GAR_NAMESPACE`?

Done.

lixueclaire (approved):
LGTM

Issue #441, opened 2024-04-07
[Spark-java-refactor] Implement file get, dump and save functions for info


Thespica:
**Is your feature request related to a problem? Please describe.**
As titile.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #443, opened 2024-04-07
[Spark-java-refactor] Implement file get, dump and save functions for info

Thespica:
## Proposed changes

Main changes:
- prefix and file get methods for info;
- dump and save methods for info;
- dump-related methods for yamlParser;

## Checklist

- [x] I have signed the CLA

## Further comments

Related issue #441

### DIFF
--- a/spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
+++ b/spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
@@ -118,6 +118,10 @@ public class EdgeInfo {
         this.version = version;
     }
 
+    public static EdgeInfo load(String edgeInfoPath) throws IOException {
+        return load(edgeInfoPath, new Configuration());
+    }
+
     public static EdgeInfo load(String edgeInfoPath, Configuration conf) throws IOException {
         if (conf == null) {
             conf = new Configuration();
--- a/spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
+++ b/spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
@@ -19,10 +19,6 @@
 
 package org.apache.graphar.info;
 
-import org.apache.graphar.info.type.AdjListType;
-import org.apache.graphar.info.type.DataType;
-import org.apache.graphar.info.yaml.EdgeYamlParser;
-import org.apache.graphar.util.GeneralParams;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
@@ -30,9 +26,14 @@ import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
+import org.apache.graphar.info.type.AdjListType;
+import org.apache.graphar.info.type.DataType;
+import org.apache.graphar.info.yaml.EdgeYamlParser;
+import org.apache.graphar.info.yaml.GraphYamlParser;
+import org.apache.graphar.util.GeneralParams;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.yaml.snakeyaml.LoaderOptions;
@@ -121,9 +122,14 @@ public class EdgeInfo {
         if (conf == null) {
             conf = new Configuration();
         }
-        Path path = new Path(edgeInfoPath);
-        FileSystem fileSystem = path.getFileSystem(conf);
-        FSDataInputStream inputStream = fileSystem.open(path);
+        return load(edgeInfoPath, FileSystem.get(conf));
+    }
+
+    public static EdgeInfo load(String edgeInfoPath, FileSystem fileSystem) throws IOException {
+        if (fileSystem == null) {
+            fileSystem = FileSystem.get(new Configuration());
+        }
+        FSDataInputStream inputStream = fileSystem.open(new Path(edgeInfoPath));
         Yaml edgeInfoYamlLoader =
                 new Yaml(new Constructor(EdgeYamlParser.class, new LoaderOptions()));
         EdgeYamlParser edgeInfoYaml = edgeInfoYamlLoader.load(inputStream);
@@ -184,6 +190,9 @@ public class EdgeInfo {
     }
 
     public AdjacentList getAdjacentList(AdjListType adjListType) {
+        // AdjListType will be checked in this method,
+        // other methods which get adjacent list in this class should call this method first,
+        // so we don't check AdjListType in other methods.
         checkAdjListTypeExist(adjListType);
         return adjacentLists.get(adjListType);
     }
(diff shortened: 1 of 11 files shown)


spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -121,9 +122,14 @@ public static EdgeInfo load(String edgeInfoPath, Configuration conf) throws IOEx
         if (conf == null) {
             conf = new Configuration();
         }
-        Path path = new Path(edgeInfoPath);
-        FileSystem fileSystem = path.getFileSystem(conf);
-        FSDataInputStream inputStream = fileSystem.open(path);
+        return load(edgeInfoPath, FileSystem.get(conf));
+    }
+
+    public static EdgeInfo load(String edgeInfoPath, FileSystem fileSystem) throws IOException {
```
acezen, line 128: Maybe the filesystem could be a optional parameter, here it seems that user need to always pass a fileSystem with this method?

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -196,98 +205,73 @@ public PropertyGroup getPropertyGroup(String property) {
...
-    //    /**
-    //     * Returns the edge info as a YAML formatted string.
-    //     *
-    //     * @return A Result object containing the YAML string, or a Status object
-    //     * indicating an error.
-    //     */
-    //    public String dump() {
-    //
-    //    }
+    public void save(String filePath, Configuration conf) throws IOException {
```
acezen, line 254: Ditto, the Configuration should be an optional parameter with default is null

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -196,98 +205,73 @@ public PropertyGroup getPropertyGroup(String property) {
...
+        if (conf == null) {
+            conf = new Configuration();
+        }
+        save(filePath, FileSystem.get(conf));
+    }
+
+    public void save(String fileName, FileSystem fileSystem) throws IOException {
```
acezen, line 261: Ditto.

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
```diff
@@ -95,8 +96,8 @@ private GraphInfo(
         this.version = version;
     }
 
-    public static GraphInfo load(String graphPath) throws IOException {
-        return load(graphPath, null);
+    public static GraphInfo load(String graphPath, FileSystem fileSystem) throws IOException {
```
acezen, line 99: Ditto.

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
```diff
@@ -112,16 +113,29 @@ public static GraphInfo load(String graphPath, Configuration conf) throws IOExce
...
 
-    // TODO(@Thespica): Implement save and dump methods
-    //    public void save(String path) {
-    //
-    //    }
-    //
-    //    public String dump() {
-    //
-    //    }
+    public void save(String filePath, Configuration conf) throws IOException {
```
acezen, line 116: Ditto.

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
```diff
@@ -112,16 +113,29 @@ public static GraphInfo load(String graphPath, Configuration conf) throws IOExce
...
+        if (conf == null) {
+            conf = new Configuration();
+        }
+        save(filePath, FileSystem.get(conf));
+    }
+
+    public void save(String fileName, FileSystem fileSystem) throws IOException {
```
acezen, line 123: Ditto.

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
Thespica: FileSystem is optional for conveinience, user can also call with hdfs' configuration object.

acezen: sorry, I don't get the meaning, you mean the fileSystem can be a FileSystem object or Configuration? Is that possible with single function in Java?

acezen: For the users that are not familiar with Hadoop's FileSystem, I think we should provide a default load function that directly load file from local filesystem. Use `XxxInfo.load(infoPath, null)` is really weird to me, since pass a null we should raise an error, not create a default filesystem in most cases.

acezen: how about we simply the load function by provide two overload function:
load(String)  # load info from create filesystem with new Configuration();
load(String, Configuration conf) # load info from filesystem create with given configuration

This would give both direct and simple cognition for users that familiar with Hadoop's FileSystem API or not familiar with the API.

Thespica: added

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
Thespica: fixed

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
Thespica: fixed

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
Thespica: fixed

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
Thespica: fixed

spark/graphar/src/main/java/org/apache/graphar/info/GraphInfo.java
Thespica: fixed

spark/graphar/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -117,13 +118,22 @@ private EdgeInfo(
...
             conf = new Configuration();
         }
-        Path path = new Path(edgeInfoPath);
-        FileSystem fileSystem = path.getFileSystem(conf);
-        FSDataInputStream inputStream = fileSystem.open(path);
+        return load(edgeInfoPath, FileSystem.get(conf));
+    }
+
+    public static EdgeInfo load(String edgeInfoPath, FileSystem fileSystem) throws IOException {
+        if (fileSystem == null) {
```
acezen, line 133: I think we should raise error if the given Configuration or FileSystem is null. If we use a default `new Configuration()` and can not read the path(which failed like user give a null but path is an oss path), it can not tell that the reason of read failed is that user pass a invalid Configuration/FileSystem.

Thespica: fixed

acezen (approved):
LGTM

Issue #454, opened 2024-04-18
[Feat][C++] Use `#pragma once` to replace include guard


acezen:
**Is your feature request related to a problem? Please describe.**
[`#pragma once`](https://en.wikipedia.org/wiki/Pragma_once) serves the same purpose as include guard but with several advantages, including less code, avoidance of name clashes, less prone to making mistakes.

**Describe the solution you'd like**
Replace include guard with `program once`.

acezen:
close as resoved

PR #455, opened 2024-04-18
[FEAT][C++] Use `#pragma once` to replace `#ifndef` as  include guards

acezen:
## Proposed changes
This change use `#pragma once` to replace `#ifndef` as  include guards

## Further comments
Fixs #454


lixueclaire (approved):
LGTM

PR #456, opened 2024-04-18
[BugFix][C++] Cast to correct schema when get chunk with property reader

### LINKED ISSUE #219: [Bug]: DataType inconsistent for Int32 in CSV file format
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Defaultly, [arrow reader for CSV](https://arrow.apache.org/docs/cpp/csv.html#data-types) does not support Int32, it will infer the data as Int64. After that, if we get vertex/edge property, which contains convert value from Int64 to Int32, it will cause an error.

### Expected Behavior

Conduct data type conversion automatically.

### Minimal Reproducible Example

```
label: person
chunk_size: 2
prefix: vertex/person/
property_groups:
  - properties:
      - name: age
        data_type: int32
        is_primary: false
    prefix: age/
    file_type: csv
version: gar/v1
```

Then, get property value of age through `VerticesCollection`.

### Environment

- Operating system: Mac OS
- GraphAr version: gar/v1

### Link to GraphAr Logs

_No response_

### Further Information

_No response_

acezen:
## Proposed changes
as issue #219 describe, the chunk table get from arrow chunk reader may not has the same type with schema.
This change help to fix the problem by cast the chunk table to what they should be(schema), if cast failed, raise error.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/apache/incubator-graphar/blob/main/CONTRIBUTING.rst) doc
- [x] I have signed the CLA

## Further comments
fixes #219

### DIFF
--- a/cpp/include/gar/reader/arrow_chunk_reader.h
+++ b/cpp/include/gar/reader/arrow_chunk_reader.h
@@ -31,6 +31,7 @@
 // forward declaration
 namespace arrow {
 class Array;
+class Schema;
 class Table;
 }  // namespace arrow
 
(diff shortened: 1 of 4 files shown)


lixueclaire (approved):
LGTM

Issue #459, opened 2024-04-22
feat(c++): Find a better uri parser for GraphAr, remove the rely on arrow::internal::URI


acezen:
**Is your feature request related to a problem? Please describe.**
currently, we use `arrow::internal::URI` to parse and check the uri string, but it is part of the Apache Arrow C++ library's internal API, which is not intended for direct use by end-users. The internal API is subject to change without notice, and its use is not recommended for production code.

We should find a third-party URI parse library to replace the `arrow::internal::URI`.

**Describe the solution you'd like**
~~Since GraphAr has already rely on Boost library. To avoid include other dependency, it's better to use `Boost::URL`.~~

Boost::URL in not available in boost library before 2023, considering use a light-weight single header parser like [simple-uri-parser](https://github.com/jholloc/simple-uri-parser)

PR #460, opened 2024-04-23
feat(c++):  Use simple-uri-parser as uri parser, remove the rely on arrow::internal::URI

acezen:
## Proposed changes
as #459 describe


lixueclaire (approved):
LGTM

Issue #479, opened 2024-05-11
fix(license): fix the LICENSE and NOTICE according to ASF requirement


acezen:
### Describe the bug, including details regarding any error messages, version, and platform.

In the infrastructure building, we have followed [Fury] to organize the LICENSE and NOTICE files: GraphAr puts  the 3-party dependencies license to NOTICE.

But according to the [infrastructure guide](https://infra.apache.org/licensing-howto.html#mod-notice) and discussion on the [general@incaubator.a.o](https://lists.apache.org/thread/g20209w7nc1qb5clfvz7wt5jv0ck3vny), all license information should goes to LICENSE file, notice information goes to NOTICE file.

### Component(s)

Other

PR #480, opened 2024-05-11
fix(license): fix LICENSE and NOTICE file according to ASF policy

acezen:
### Reason for this PR

 as issue #479 describes

for 3rd-party dependencies of graphar:
- put licence information  to LICENSE
- put notice information  to NOTICE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,83 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+================================================================
+This product includes a number of dependencies with separate license terms.
+Your use of these submodules is subject to the terms and conditions of the 
+following licenses.
+================================================================
+
+================================================================
+Apache-2.0 licenses
+================================================================
+The following components are provided under the Apache-2.0 License. See project link for details.
+The text of each license is the standard Apache 2.0 license.
+
+* spark 3.1.1 and 3.3.4 (https://github.com/apache/spark)
+    Files:
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarCommitProtocol.scala
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarDataSource.scala
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarScan.scala
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarScanBuilder.scala
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarScanTable.scala
+      maven-projects/spark/datasourcs-32/src/main/scala/org/apache/graphar/datasources/GarWriteBuilder.scala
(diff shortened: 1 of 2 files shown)


acezen:
Hi, @CalvinKirs @carp84, could you help review this PR?

NOTICE
```diff
@@ -4,86 +4,47 @@ Copyright 2024 The Apache Software Foundation
...
+which includes the following in its NOTICE file:
+
+  GraphScope
+  Copyright 2020 Alibaba Group Holding Limited. 
+
+--------------------------------------------------------------------------------
+
+This product includes code from https://github.com/alibaba/fastFFI (Apache 2.0)
+  * Copyright 1999-2021 Alibaba Group Holding Ltd.
+
```
CalvinKirs, line 31: If the third-party code does not have a corresponding NOTICE file, it is not necessary to add one.

NOTICE
```diff
@@ -4,86 +4,47 @@ Copyright 2024 The Apache Software Foundation
...
+--------------------------------------------------------------------------------
+
+This product includes code from https://github.com/jimmiebergmann/mini-yaml (The MIT License)
+  * Copyright (c) 2021 Jonathan Hollocombe
+
+--------------------------------------------------------------------------------
+
+This product includes code from cpplint (https://github.com/cpplint/cpplint) (BSD 3-Clause)
+  * Copyright (c) 2009 Google Inc. All rights reserved.
```
CalvinKirs, line 50: MIT and BSD do not require NOTICE files, and they typically don't include NOTICE files.

LICENSE
```diff
@@ -199,3 +199,83 @@
...
+
+================================================================
+MIT licenses
+================================================================
+The following components are provided under the MIT License. See project link for details.
+The text of each license is also included in licenses/LICENSE-[project].txt
+
+* result (https://github.com/bitwizeshift/result)
+    Files:
+      cpp/thirdparty/result.hpp is modified from result.
```
CalvinKirs, line 262: It would be even better if we could include the version of the corresponding third-party code, licenses may change over time.

NOTICE
acezen: Thanks for pointing out, already removed the third-party dependency without NOTICE file form NOTICE

CalvinKirs:
Overall, this looks great! I noticed that we also kept the corresponding license header, which is excellent. I haven't checked if the referenced code has been modified on top of it, but if so, we can also add the corresponding AL2 header. FYI https://issues.apache.org/jira/browse/LEGAL-626

LICENSE
acezen: Is that ok that `mini-yaml` and `simple-uri-parser` provide a commit hash since they did not release any version by now?

acezen:
> Overall, this looks great! I noticed that we also kept the corresponding license header, which is excellent. I haven't checked if the referenced code has been modified on top of it, but if so, we can also add the corresponding AL2 header. FYI https://issues.apache.org/jira/browse/LEGAL-626

Thanks for the reminder, some of them  modified on top of the origin code. I would check and submit another PR to add the corresponding AL2 header.

LICENSE
CalvinKirs: absolutely ok:)

lixueclaire (approved):
LGTM

CalvinKirs (approved):

Issue #478, opened 2024-05-10
[java-info] Merge java-info code into main by components step by step


Thespica:
### Describe the enhancement requested

Now, the java-info code basically has been completed in [this branch](https://github.com/apache/incubator-graphar/tree/branch-feat-spark-java-info-refactoring). I am going to merge these code into main branch by components step by step:
- [x] basic architecture(including pom.xml and directory), yaml, type and util
- [x] small class like Property and AdjacentList
- [x] graph, vertex and edge info

### Component(s)

Format, Java, Spark

SemyonSinchenko:
May we postpone that until we decide with proto?
(see #475 and discussion in mailing list)

Thespica:
Hi @SemyonSinchenko , I think this two missions are relatively independent, the merge won't effect the decision about proto.

acezen:
> May we postpone that until we decide with proto? (see #475 and discussion in mailing list)

I think is ok to migrate the current implementation from branch-feat-spark-java-info-refactoring to main. when the discussion about protobuf is done, we can refactoring the module base the main.

PR #481, opened 2024-05-14
[java-info] Add basic architecture java-info into main

Thespica:
### Reason for this PR

As we discussed in issue #478 , this PR aims to work for the first step.

Basic architecture(including pom.xml and directory), yaml, type and util

### Are these changes tested?

no

no


maven-projects/java-info/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.graphar</groupId>
+        <artifactId>graphar-root</artifactId>
+        <version>${graphar.version}</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>java-info</artifactId>
```
acezen, line 35: It's better to use `info` as package and directory name since spark and java reader/writer would use the module directly

maven-projects/pom.xml
```diff
@@ -32,6 +32,7 @@
   <modules>
     <module>java</module>
     <module>spark</module>
+    <module>java-info</module>
```
acezen, line 35: Ditto

acezen:
Overall LGTM, we can change the directory and package name to `info`

acezen (changes requested):
Overall LGTM, we can change the directory and package name to `info`

maven-projects/java-info/pom.xml
Thespica: Updated, please review again.

acezen (approved):
LGTM

PR #482, opened 2024-05-15
feat(java): Set parent pom to apache

CalvinKirs:
### Reason for this PR

https://maven.apache.org/pom/asf/
Apache Software Foundation Parent POM
contains settings that are likely to be useful to any Apache project that is building and releasing code with Maven. By using this project as a parent, a project gets these settings.
**If we have specific version requirements for certain plugins, we need to explicitly declare the corresponding versions in the project. Otherwise, it will default to the version specified in the parent POM.**
- Add mailing list and License etc

### DIFF
--- a/maven-projects/pom.xml
+++ b/maven-projects/pom.xml
@@ -22,20 +22,58 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache</groupId>
+        <artifactId>apache</artifactId>
+        <version>31</version>
+    </parent>
+    <groupId>org.apache.graphar</groupId>
+    <artifactId>graphar-root</artifactId>
+    <version>${graphar.version}</version>
+    <packaging>pom</packaging>
 
-  <groupId>org.apache.graphar</groupId>
-  <artifactId>graphar-root</artifactId>
-  <version>${graphar.version}</version>
-  <packaging>pom</packaging>
+    <name>Apache GraphAr Root POM</name>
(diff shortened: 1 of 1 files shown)


maven-projects/pom.xml
```diff
@@ -22,20 +22,58 @@
...
+<!--    <issueManagement>
+        <system>GitHub</system>
+        <url>https://github.com/apache/graphar/issues</url>
+    </issueManagement>
+    <scm>
+        <connection>scm:git:https://github.com/apache/graphar.git</connection>
+        <developerConnection>scm:git:https://github.com/apache/graphar.git</developerConnection>
+        <tag>HEAD</tag>
+        <url>https://github.com/apache/graphar</url>
+    </scm>-->
```
CalvinKirs, line 70: https://lists.apache.org/thread/1frk5f146pbzprj0d1ftf15o4z6v3ocs  Do we have any plans to drop the "incubator" prefix? There's been an email thread discussing this in the incubator.

lixueclaire (approved):
LGTM. Thanks for making this change!

maven-projects/pom.xml
lixueclaire: > https://lists.apache.org/thread/1frk5f146pbzprj0d1ftf15o4z6v3ocs Do we have any plans to drop the "incubator" prefix? There's been an email thread discussing this in the incubator.

Thanks for bringing this information to us. Are there specified steps or guidelines that we should follow in response to this?

CalvinKirs: Typically, related to infrastructure needs to be handled by the INFRA team. If we're planning to drop the "incubator" prefix, we'll just need to submit a ticket to the INFRA JIRA.
https://issues.apache.org/jira/projects/INFRA/summary

lixueclaire: > Typically, related to infrastructure needs to be handled by the INFRA team. If we're planning to drop the "incubator" prefix, we'll just need to submit a ticket to the INFRA JIRA. https://issues.apache.org/jira/projects/INFRA/summary

Thanks! I'll consult with other members and make a decision.

acezen (approved):
LGTM, thanks~

PR #484, opened 2024-05-17
[java-info] Add vertex, edge, graph info

Thespica:
### Reason for this PR

The third step in issue #478 .

- Add vertex, edge, graph info;
- Move AdjList, Property, PropertyGroup into info package.

### Are these changes tested?

no

no


acezen (approved):
Hi, John, you may need to add some API document in the code , that would help user/developer to know what the method use for

SemyonSinchenko:
> Hi, John, you may need to add some API document in the code , that would help user/developer to know what the method use for

Is it the right time for that? Because we are in the middle of a decision about using proto: with proto info, classes should be refactored again. I would like to recommend postponing writing docs.

acezen:
> > Hi, John, you may need to add some API document in the code , that would help user/developer to know what the method use for
> 
> Is it the right time for that? Because we are in the middle of a decision about using proto: with proto info, classes should be refactored again. I would like to recommend postponing writing docs.

Yes, you are right, considering the proto decision, postponing the docs avoids refactor again, thanks Sem

PR #485, opened 2024-05-20
fix(license): add reference code to java related third-party code

acezen:
### Reason for this PR

as apache [legal discussion](https://issues.apache.org/jira/browse/LEGAL-626) describe, if the code is modified on top of the reference code, we can also add the corresponding AL2 header.

add the corresponding AL2 header to reference code if modified on top of it.

### Are these changes tested?

no need to test

no


lixueclaire (approved):
LGTM

PR #507, opened 2024-05-29
feat(dev): Add release and verify scripts

acezen:


Thespica (approved):
LGTM, but I haven't verified it locally.

Issue #510, opened 2024-05-31
feat(CI): Remove the testing submodule and use a specified environment variable to indicate the testing data directory


acezen:
### Describe the enhancement requested

To not include the testing data into source tarball and still make the testing work, it's better to not make testing data as a submodule and include it with a outer directory.

### Component(s)

Continuous Integration

PR #512, opened 2024-06-03
Feat(CI): Use  environment variable to specify the location of testing data

acezen:
### Reason for this PR

as issue #510 describes, since the released source code tarball would not contains the submodules, to make the test work with tarball, it's better to deprecate testing data submodule and use developer setting environment variable to indicate the location of testing data.

- refine the unit test of c++/scala library with the 'GAR_TEST_DATA' env variable
- update the related developer document 
- [TODO] Drop the testing submodule until java/python library update the unit test like cpp/scala. (java/python not bump up in this release)

### Are these changes tested?

already tested by current test case.

no


acezen:
Hi， @lixueclaire @SemyonSinchenko ， This PR may need your review, thanks.

lixueclaire (approved):
LGTM

cpp/README.md
```diff
@@ -88,21 +88,27 @@ Debug build with unit tests:
...
     $ cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON ..
     $ make -j8       # if you have 8 CPU cores, otherwise adjust, use -j`nproc` for all cores
-    $ make test      # to run the tests
+```
+
+After building, you can run the unit tests with:
+
+```bash
+    $ git clone https://github.com/apache/incubator-graphar-testing.git testing  # download the testing data
+    $ GAR_TEST_DATA=${PWD}/testing ctest
```
SemyonSinchenko, line 97: I would recommend to create a folder `dev/` and provide bash/python scripts for downloading the data and set up all the variables. We can do it in a separate PR. It is done like this in [spark](https://github.com/apache/spark/tree/master/dev) and I found it very user-friendly.

SemyonSinchenko (approved):

cpp/README.md
acezen: The suggestion is great, I can do it in the PR #507 .

PR #514, opened 2024-06-05
feat(spark): Refactoring datasources

### LINKED ISSUE #493: [Spark] Move datasources under org.apache.spark package

Currently datasources implementation are under `org.apache.graphar` space. I suggest to move these classes to `org.apache.spark` package. It allows us, for example, to use spark internals. For example, it resolve the problem in #488 when `JSONOptions` is `private[sql]`.

### Component(s)

Spark, PySpark

SemyonSinchenko:
### Reason for this PR
By moving datasources under `org.apache.spark.sql` we are able to access private Spark API. Last time when I was trying to fully migrate datasources to V2 it was a blocker. Detailed motivation is in #493 

Mostly refactoring.

### Are these changes tested?
Unit tests are passed

I manually checked the generated JARs:
image

Mostly not because `GarDataSource` was left under the same package.

Close #493


SemyonSinchenko:
@acezen What do you think about this? For me it looks like there are no changes for public API

acezen:
> @acezen What do you think about this? For me it looks like there are no changes for public API

Hi, Sem, thanks for the draft, change to `org.apache.spark.sql` is good to me.  But if we upload the binary package to maven, is that possible to upload all the binary packages to `org.apache.graphar`?

acezen:
> > > @acezen What do you think about this? For me it looks like there are no changes for public API
> > 
> > 
> > Hi, Sem, thanks for the draft, change to `org.apache.spark.sql` is good to me. But if we upload the binary package to maven, is that possible to upload all the binary packages to `org.apache.graphar`?
> 
> Yes, it should be. For example, check this:
> 
> 1. Download delta-saprk: https://mvnrepository.com/artifact/io.delta/delta-spark_2.13/3.2.0
> 2. Open it with archive manager
> 3. You will see org-apache-spark classes inside:
>    image
> 
> but coordinates are still `io.delta:delta-spark_2.13:3.2.0`
> 
> So, if you OK with it I will update 32 too and fix license headers.

Looks NICE! The change is OK for me.  Thanks for the work.

SemyonSinchenko:
Ready for review.

acezen (approved):
LGTM

PR #521, opened 2024-06-13
feat(format): Implement the protocol definition of GraphAr format

acezen:
### Reason for this PR

After implementation and including buf to graphar format definition, it's time to merge the format develop branch in to main.

- implementation of protocol of GraphAr format definition, there are some place that diff from the libraries:
    - the `label` naming to `type`, to distinguish with the multi-label support in the future.
    - extend the metadata to the definition, to including the number of vertex/edge, the chunk number of vertex, the edge num of each vertex chunk etc. We recorded the metadata with many files before.
- Use buf to generate code from protobuf.
- Add related CI

No


SemyonSinchenko:
Why diff is so big? May you, please, make a force-push from main to the branch?

format/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Prerequisites
+
+- [protoc](https://developers.google.com/protocol-buffers/docs/downloads)
+- [buf](https://buf.build/docs/installation)
+
+### Build
+
+```sh
+buf generate
```
SemyonSinchenko, line 15: It may be little confusing because one needs to run this command from the root of the project, not from the `format` folder

format/README.md
```diff
@@ -0,0 +1,18 @@
+# GraphAr Format Specification
+
+This folder contains protocol definitions for the GraphAr format.
+
+## How to generate code
+
+### Prerequisites
+
+- [protoc](https://developers.google.com/protocol-buffers/docs/downloads)
+- [buf](https://buf.build/docs/installation)
```
SemyonSinchenko, line 10: It is important that we need to use buf >= 1.32 because we are on v2. May we mention it here?

format/test_vertex_info.cc
```diff
@@ -0,0 +1,57 @@
+/*
```
SemyonSinchenko, line 1: What is the reason of that file?

format/README.md
acezen: Added the version tag, thanks~

format/README.md
acezen: I have added some wording to indicate that the command should run it the root of repository.

format/test_vertex_info.cc
acezen: Sorry, this is the test file of POC, already removed it.

SemyonSinchenko (approved):

SemyonSinchenko:
By the way, based on implemented GH action I can assume that we are not storing generated code in git, right? Did you see my email in a list about it? For me the question is not so clear. @acezen

acezen:
> By the way, based on implemented GH action I can assume that we are not storing generated code in git, right? Did you see my email in a list about it? For me the question is not so clear. @acezen

Yes, I see the disscussion. Currently we do not storing generated code. But if some libraries implementation prefer to storing generated code, it can be added in the future.

Issue #522, opened 2024-06-14
[Feat][Format][Java][Spark] Incorporate maven-proto-plugin into build of Maven projects


SemyonSinchenko:
### Describe the enhancement requested

Based on the discussion in a dev-list it should be a separate subproject.

### Component(s)

Format, Java, Spark

Thespica:
Hi @SemyonSinchenko 

I am available recently, wanting to help incorporating the java-info part.

 Have you finished the part of java-info? If you haven't start for this part, can you assign it to me?

PR #535, opened 2024-07-07
[Maven-project] Incorporate maven-proto-plugin into build of Maven projects

Thespica:
### Reason for this PR

See #522 

Add maven-proto-plugin into pom.xml of java-info.

When the maven project is built, this plugin will compile protobuf into java source code into `java-info/src/main/java/org/apache/graphar/proto`.

Potential problem:

 1. Are generated-code located in  `java-info/src/main/java/org/apache/graphar/proto` suitable?
 2. Once the code was generated, developer should remove the code manually to avoid compile error.

### Are these changes tested?

No


SemyonSinchenko:
If protobuf is in maven build already let's remove it from `buf.gen.yaml`: https://github.com/apache/incubator-graphar/blob/main/buf.gen.yaml#L33

maven-projects/info/pom.xml
```diff
@@ -42,6 +42,9 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <os.plugin.version>1.6.2</os.plugin.version>
+        <protobuf.plugin.version>0.6.1</protobuf.plugin.version>
+        <protobuf.version>3.21.1</protobuf.version>
```
SemyonSinchenko, line 47: I would suggest to pin the version not only for Maven builds but globally for the whole GraphAr. Otherwise we may face problem one day if we will try to serialize from java and, for example, read from cpp

maven-projects/info/pom.xml
```diff
@@ -63,5 +66,37 @@
...
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>${os.plugin.version}</version>
+            </extension>
+        </extensions>
+        <plugins>
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
```
SemyonSinchenko, line 79: While this `xolostice` plugin is more popular at the moment, I like this one more: https://github.com/ascopes/protobuf-maven-plugin

I found it more human-friendly and easier to work with.

SemyonSinchenko: There is a good motivation: https://github.com/ascopes/protobuf-maven-plugin?tab=readme-ov-file#why-do-we-need-another-plugin

SemyonSinchenko: With this plugin we do not need to have `os-plugin`

acezen:
Maybe we can put the location of generated code to `maven-projects/src/main/java/org/apache/graphar/proto/`? since we want the generated code can be a independent package like [arrow java's flatbuf](https://github.com/apache/arrow/tree/main/java/format)

acezen:
Why it would raise compile error if not delete the generated code?

maven-projects/info/pom.xml
Thespica: Yes, we can follow our [doc](https://github.com/apache/incubator-graphar/blob/main/format/README.md) and pin it at 1.32.0

SemyonSinchenko: 1.32.0 is the version of buf, not `protoc`.

Th actual version of protobuf that we are using is `3.27.1`
image

(explanation: we are on `proto3` syntax and the version of protobuf plugin for a minor version of `protoc` itself)

I mean, I have a strong feeling, that we need to pin it globally. Otherwise it may become hard to maintain it in two different places (Maven and buf.gen.yaml)

acezen: > 1.32.0 is the version of buf, not `protoc`.
> 
> Th actual version of protobuf that we are using is `3.27.1` image
> 
> (explanation: we are on `proto3` syntax and the version of protobuf plugin for a minor version of `protoc` itself)
> 
> I mean, I have a strong feeling, that we need to pin it globally. Otherwise it may become hard to maintain it in two different places (Maven and buf.gen.yaml)

agree

Thespica:
> maven-projects/src/main/java/org/apache/graphar/proto/

With modifications, the code well generated at `maven-projects/proto/src/main/java/org/apache/graphar/proto`, and will be clean automatically when regenerating.

SemyonSinchenko:
Guys, why do we need both, buf-java plugin and maven-protobuf plugin? For me, if we do not want to store the code in the repository, let's remove java-generation from buf and leave only the maven plugin. Or did I miss something?

maven-projects/info/pom.xml
SemyonSinchenko: The current protoc version is **3.27.1**; at least we may pin it in Maven properties

If we are not going to use buf for generatign Java code, we do not need to touch buf version from this PR

acezen:
> Guys, why do we need both, buf-java plugin and maven-protobuf plugin? For me, if we do not want to store the code in the repository, let's remove java-generation from buf and leave only the maven plugin. Or did I miss something?
hi, Sem, you are not miss anything. We do need leave only the maven plugin. @Thespica, could you delete the [java-generation](https://github.com/apache/incubator-graphar/blob/759c6219ca9d39bc687bffceba6da11f124cc38c/buf.gen.yaml#L33-L35) from buf if we use maven plugin.

maven-projects/info/pom.xml
Thespica: This plugin requires maven 3.8.2, which means we have to upgrade the requirement of maven version if we accept the newer plugin.

image

acezen:
> > maven-projects/src/main/java/org/apache/graphar/proto/
> 
> With modifications, the code well generated at `maven-projects/proto/src/main/java/org/apache/graphar/proto`, and will be clean automatically when regenerating.

Sorry for spell the path wrong, I suggest we can put the generated code to `maven-projects/format/src/main/java/org/apache/graphar/proto`, since it's the generated code for format.

Thespica:
> > > maven-projects/src/main/java/org/apache/graphar/proto/
> > 
> > 
> > With modifications, the code well generated at `maven-projects/proto/src/main/java/org/apache/graphar/proto`, and will be clean automatically when regenerating.
> 
> Sorry for spell the path wrong, I suggest we can put the generated code to `maven-projects/format/src/main/java/org/apache/graphar/proto`, since it's the generated code for format.

I think it's okay to naming module as `proto`, since the package name is proto.

acezen (approved):
I have checked the compilation. LGTM, can you add compile process to format CI?

maven-projects/info/pom.xml
SemyonSinchenko: I would update the Maven. I see no reason to keep older version without a reason

Thespica:
Hi @SemyonSinchenko 

I have transfered to the newer plugin, could you review again?

SemyonSinchenko:
It looks cool. Thank you!

I still have some concerns. It seems to me that we should use a maven-shade plugin to shade protobuf and rename generated classes... Otherwise, users may face dependency hell. But it is out of scoop for the current PR. Let's go step by step.

.github/workflows/format.yml
```diff
@@ -56,4 +56,10 @@ jobs:
 
       - name: Buf Generate
         run: buf generate
- 
\ No newline at end of file
+
+      - name: Java Code Generate
+        run: |
+          pushd maven-projects/proto
+          mvn clean compile
```
SemyonSinchenko, line 63: [Minor] why compile, not `generate-sources`?

SemyonSinchenko (approved):

Issue #549, opened 2024-07-24
bug(c++): arrow reader corner case


jasinliu:
### Describe the enhancement requested

When we use reader->GetChunk(), if the directory starts with part1, we can not get the chunk. But next_chunk() is fine. If we don’t know the directory, we don't know wheher we need use next_chunk().

Example: neo4j/edge/Person_REVIEWED_Movie/ordered_by_source/adj_list/**part1**.
https://github.com/apache/incubator-graphar-testing/tree/main/neo4j/edge/Person_REVIEWED_Movie/ordered_by_source

### Component(s)

C++

acezen:
Thanks for the report, I will take care of this bug.

jasinliu:
Minimal reproduction process https://gist.github.com/jasinliu/e7d20688fcbc17de22fb3b6fa13eb59d

![QQ_1721820477740](https://github.com/user-attachments/assets/78e70db2-44e6-4a1f-b416-167f848278ab)

jasinliu:
Oh, this is my question. 

We need to know a premise: the first chunk can be empty. Then make a judgment through `has_value()` to decide whether to switch to the next chunk.

acezen:
Reopen since we suppose to return nullptr if the edge of the vertex chunk is empty.

acezen:
done

PR #550, opened 2024-07-25
fix(c++): return nullptr at GetChunk if there is no edge in chunk

acezen:
### Reason for this PR

as #549 describe, when there are no edge of a vertex chunk, the directory could be empty, and the GetChunk would return `IOError`. 

Instead of return IOError, change to return a nullptr since the GetChunk operation should be OK and just the return chunk is empty.

### Are these changes tested?

the test case will be added.

No


acezen:
Hi, @jasinliu, thanks for the bug report, this is the fix for the bug, could you help to take a review?

lixueclaire (approved):
LGTM

jasinliu (commented):
LGTM. Thanks very much~

Issue #547, opened 2024-07-23
Dependency arrow/orc not installed by brew on macOS


yecol:
### Describe the bug, including details regarding any error messages, version, and platform.

It seems we have recognized this problem as shown in the cpp/README.md.
> Note: Currently, the Arrow C++ library is [disable ARROW_ORC](https://github.com/Homebrew/homebrew-core/blob/4588359b7248b07379094de5310ee7ff89afa17e/Formula/a/apache-arrow.rb#L53) in the brew formula, so you need to install the Arrow C++ library manually.

However, the given solution seems does not fix the problem. 
Since in the `cpp/Brewfile`, it still just simply `install apache-arrow`

### Component(s)

C++, Developer Tools

acezen:
We suppose to fix that by #545 #546

Issue #545, opened 2024-07-22
feat(c++): Raise error in CMakeList.txt if arrow does not support ORC


acezen:
### Describe the enhancement requested

Sometimes the installed arrow dependency does not support ORC. For example, current homebrew apache-arrow package [does not include ORC support](https://github.com/Homebrew/homebrew-core/blob/3e5bcaebc162335f428571a4cb0277fcdb35886d/Formula/a/apache-arrow.rb#L52)

We need to add a check to CMakeList, raise error if the arrow does not support ORC.

### Component(s)

C++

acezen:
The best way to check that ORC support or not is check the `ARROW_ORC` variable after `find_package(Arrow)`.

yecol:
Thanks for your hint! I will try.

PR #552, opened 2024-07-25
fix(c++): Raise error if cmake found arrow built without ORC support.

yecol:
### Reason for this PR

Since brew installed arrow does not include orc extension. This PR add a fix to raise fatal message to terminate the build process if it does not find the orc extension.
Fix issues #547 and #545 

a testing item added in CMakeList.

### Are these changes tested?

Yes. but the test case is not included in PR, since it only affects a building failed codepath.

No.


acezen (approved):
Thanks, the fix looks good to me.

PR #561, opened 2024-07-29
feat(spark): Support local filesystem in spark library for graphar-info temporarily

### LINKED ISSUE #554: feat(spark): Support local filesystem without sparksession

Curently, the scala library needs a spark process to use. But it will be a bit complicated if we only deal with local files.

I hope that we can support the local filesystem without sparksession. Now my idea is to allow the loader and some funcs in the scala library to allow the spark parameter to be None. 

```scala
def loadGraphInfo(
      graphInfoPath: String,
      spark: Option[SparkSession] = None
  )
 ```

### Component(s)

Spark

jasinliu:
### Reason for this PR

fix #554 

as title.


jasinliu:
This PR brings a very big change in the usage. When using `spark` parameter in `info`, users must add `Some` to support this `Option`.

jasinliu:
> This PR brings a very big change in the usage. When using `spark` parameter in `info`, users must add `Some` to support this `Option`.

I want to know how to avoid changing the user mode?

jasinliu:
> This PR brings a very big change in the usage. When using `spark` parameter in `info`, users must add `Some` to support this `Option`.

Before the `info` module arrives, if this PR makes sense, I can try to modify and add some tests.

SemyonSinchenko (changes requested):
GraphAr PySpark is broken by these changes.

Issue #544, opened 2024-07-22
chore(ci):  Open the macOS CI


acezen:
### Describe the enhancement requested

Since the `homebrew` has enable ARROW_ORC in apache-arrow formula, we can reopen the macOS CI.

### Component(s)

Continuous Integration

PR #558, opened 2024-07-29
feat: make apache-arrow ORC extension optional

yecol:
### Reason for this PR

To fix #546,  following the discussion.

### What changes are included in this PR?

Using `ARROW_ORC` macro to safe guard the orc related code.

### Are these changes tested?

Yes, tests have changed correspondly and passed. 

### Are there any user-facing changes?

No.

Issue #271, opened 2023-11-02
[Bug][Java]: Bridge dynamic library gar-jni build fatal with Apache Arrow 14.0.0


Thespica:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When link arrow as static:
```
[INFO]      [exec] -- Found re2Alt: /usr/lib/x86_64-linux-gnu/libre2.so  
[INFO]      [exec] -- Found BZip2: /usr/lib/x86_64-linux-gnu/libbz2.so (found version "1.0.8") 
[INFO]      [exec] -- Looking for BZ2_bzCompressInit
[INFO]      [exec] -- Looking for BZ2_bzCompressInit - found
[INFO]      [exec] -- Found utf8proc: /usr/lib/x86_64-linux-gnu/libutf8proc.so (found version "2.7.0") 
[INFO]      [exec] -- Found nlohmann_json: /usr/lib/cmake/nlohmann_json/nlohmann_jsonConfig.cmake (found version "3.10.5") 
[INFO]      [exec] -- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so (found version "7.81.0")  
[INFO]      [exec] -- Arrow version: 14.0.0
[INFO]      [exec] -- Found the Arrow shared library: /usr/lib/x86_64-linux-gnu/libarrow.so.1400.0.0
[INFO]      [exec] -- Found the Arrow import library: ARROW_IMPORT_LIB-NOTFOUND
[INFO]      [exec] -- Found the Arrow static library: /usr/lib/x86_64-linux-gnu/libarrow.a
[INFO]      [exec] -- Configuring done (2.6s)
[INFO]      [exec] -- Generating done (0.1s)
[INFO]      [exec] CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Arrow/ArrowTargets.cmake:74 (set_target_properties):
[INFO]      [exec]   The link interface of target "Arrow::arrow_static" contains:
[INFO]      [exec] 
[INFO]      [exec]     protobuf::libprotobuf
[INFO]      [exec] 
[INFO]      [exec]   but the target was not found.  Possible reasons include:
[INFO]      [exec] 
[INFO]      [exec]     * There is a typo in the target name.
[INFO]      [exec]     * A find_package call is missing for an IMPORTED target.
[INFO]      [exec]     * An ALIAS target is missing.
[INFO]      [exec] 
[INFO]      [exec] Call Stack (most recent call first):
[...]

PR #563, opened 2024-07-30
Chore(CI): Opened macOS CI

yecol:
### Reason for this PR

Fixed issue #544, since now apache-arrow ORC extension is optional (by #558)

Opened CI.

### Are these changes tested?

Tested by CI.

No.


yecol:
Adding the missing protobuf includes, which occurs in macOS systems. 
Also fix the error reported in #271

acezen (approved):
LGTM, it seems that ARM64 architecture has some problem with static arrow. I think it is ok to disable the `Use Static Arrow` step on `ARM64`  and leave a reopen TODO.

PR #568, opened 2024-07-31
docs: change images in docs from relative url to absolute url.

yecol:
### Reason for this PR

Fix an image cannot display bug when open docs from README docs. 

Currently, the images in the docs are given the relative URL, which will lead to a cannot display bug:
Since clicking from github README, the link will automatically appending a '/', e.g., 
clicking on https://graphar.apache.org/docs/specification/format, 
it actually open the URL https://graphar.apache.org/docs/specification/format/

The endling '/' affected the relative path, (by add an extra level), leading the images cannot by found and return a 404 error. 

Screenshot 2024-07-31 at 15 23 35

Using absolute path can solve this problem.
Better still, even the [docs in github repo kept correctly displayed](https://github.com/yecol/incubator-graphar/blob/docs/docs/specification/format.md).

### DIFF
--- a/docs/overview/overview.md
+++ b/docs/overview/overview.md
@@ -4,7 +4,7 @@ title: Overview
 sidebar_position: 1
 ---
 
-<img src="../images/overview.png" alt="overview" width="700"/>
+<img src="/docs/images/overview.png" alt="overview" width="700"/>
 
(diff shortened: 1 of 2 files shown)


acezen (approved):
LGTM, thanks for the fix!

Issue #543, opened 2024-07-22
chore(format): Add code generated for Go language


acezen:
### Describe the enhancement requested

In discussion #540, we can provide a basic implementation by add code generated in [`buf.gen.yaml`](https://github.com/apache/incubator-graphar/blob/main/buf.gen.yaml) for Go language.

### Component(s)

Format

PR #573, opened 2024-08-02
feat: Add configuration to generate protos for Go.

yecol:
### Reason for this PR

To address #543 

- add configuration;
- fix a typo: well -> will

### Are these changes tested?

Yes.

No.


buf.gen.yaml
```diff
@@ -30,4 +30,7 @@ plugins:
   # Cpp
   - remote: buf.build/protocolbuffers/cpp:v27.1
     out: cpp/proto
-  # Java code well be generated by maven plugin
\ No newline at end of file
+  # Go
+  - remote: buf.build/protocolbuffers/go
```
SemyonSinchenko, line 34: Should we specify the version? I already faced incompatibility between code generated by different protobuf versions

yecol: Sure. However, I am not very sure about the desired version about this plugin. 
As the buf docs, it provides `buf.build/protocolbuffers/go:v1.31.0`
Do you think this version is OK? (No need to be the same with the version number for cpp/python, i.e., 27.1, right?)

SemyonSinchenko: It should be 27.1 in my understanding. Just because it is the actual version of `protoc` compiler, that is used under the hood. We are [using](https://github.com/apache/incubator-graphar/blob/main/maven-projects/proto/pom.xml#L46) the same in Java.

yecol:
Screenshot 2024-08-03 at 10 36 42


It seems the oldest version is v1.28.1
Do you think it is an acceptable compatible version?

SemyonSinchenko:
@yecol It looks like protobuf-go uses a different versioning. We can use 1.34.2, I checked and they are saying it is based on the v27-version of `protoc`. Sorry for the confusion.

yecol:
Thanks @SemyonSinchenko, I revised it to v1.34.2.
I also drop a line of comments about the version compatibility. :)

SemyonSinchenko (approved):
Thank you for that contribution! LGTM

acezen (approved):
LGTM

PR #588, opened 2024-08-12
[format] Unify name from target to destination

### LINKED ISSUE #587: [format] Unify name from `target` to `destination`

There are naming conflict with single concept in proto:
AdjListType call it target
image
meanwhile EdgeInfo call it destination
image

We can unify them to `destination`.

### Component(s)

Format

Thespica:
### Reason for this PR

close #587 

notice that there are still a lots of `target` in scala code(or documentation), maybe we should unify them too, or remain them?

rename `target` to `destination` for AdjListType in proto and concepts.

No, cause protobuf is still under development.


SemyonSinchenko (approved):
LGTM

PR #589, opened 2024-08-13
feat(devinfra): Add  test coverage to CI to measure the covering code of unitest (Spark)

Elssky:
### Reason for this PR

solution to https://github.com/apache/incubator-graphar/issues/189, extension of https://github.com/apache/incubator-graphar/pull/580

Add Spark test coverage to CI to measure the covering code of unitest### Are these changes tested?

no

### DIFF
--- a/.github/workflows/spark.yaml
+++ b/.github/workflows/spark.yaml
@@ -92,6 +92,18 @@ jobs:
         echo "Test ${{ matrix.mvn-profile }}"
         mvn test --no-transfer-progress -Dspotless.check.skip=true -P ${{ matrix.mvn-profile }}
 
+    - name: Generate JaCoCo Coverage Report
+      working-directory: maven-projects/spark
+      run: |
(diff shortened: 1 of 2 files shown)


SemyonSinchenko:
May we extend it to PySpark too? The coverage is already computed during the CI but it doesn't published. I see, that the Codecov tool [has a **support](https://about.codecov.io/tool/pytest/)** of `pytest` that is used for PySpark.

Elssky:
> May we extend it to PySpark too? The coverage is already computed during the CI but it doesn't published. I see, that the Codecov tool [has a support](https://about.codecov.io/tool/pytest/) of `pytest` that is used for PySpark.
> 
> The entry point of PySpark tests is the Makefile: https://github.com/apache/incubator-graphar/blob/main/pyspark/Makefile#L35

Sure, I'll keep track of this task👌

acezen (approved):
LGTM

PR #592, opened 2024-08-13
chore(spark,java): extract format plugin into parent module

### LINKED ISSUE #591: [maven-projects] extract format plugin into parent module

We use spotless plugin to format our java code and scala code.

currently, spotless are used in spark module and java module. In fact, we can extract it to root module to share it through submodule.

related link:
- https://stackoverflow.com/questions/10483180/what-is-pluginmanagement-in-mavens-pom-xml
- https://www.baeldung.com/maven-plugin-management

### Component(s)

Java, Spark

Thespica:
### Reason for this PR

close #591 

reorganized the spotless maven plugin.

### Are these changes tested?

yes

no

### DIFF
--- a/maven-projects/info/pom.xml
+++ b/maven-projects/info/pom.xml
@@ -63,4 +63,22 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>${spotless-maven-plugin.version}</version>
+                <configuration>
+                    <java>
+                        <googleJavaFormat>
+                            <version>1.7</version>
+                            <style>AOSP</style>
+                        </googleJavaFormat>
+                    </java>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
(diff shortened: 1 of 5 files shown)


maven-projects/info/pom.xml
```diff
@@ -63,4 +63,22 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>${spotless-maven-plugin.version}</version>
+                <configuration>
```
SemyonSinchenko, line 72: Is it possible to inherit this plugin from the parent instead of defining it explicitly?

acezen (commented):
As Sem suggested, can we put the dependency to the root pom and the modules just inherit it from parent?

Thespica:
> As Sem suggested, can we put the dependency to the root pom and the modules just inherit it from parent?

From my survey, use `plugins` can work to submodule directly, but `pluginsManagement` works better to manage them.

And I have tried use `plugins` without management directly, it's hard to work.

https://github.com/apache/avro/blob/8026c8ffe4ef67ab419dba73910636bf2c1a691c/lang/java/pom.xml#L307-L334

PR #601, opened 2024-08-21
feat(devinfra): Add test coverage to CI to measure the covering code of unitest (PySpark)

Elssky:


SemyonSinchenko (approved):
LGTM, thanks for that contribution!

Elssky:
Now there is a problem: we have three folders: pyspark, spark, cpp need to perform code coverage detection, and add to the three corresponding coverage yml for CI detection and upload to the codecov website, but in most cases every time we Only modify one folder, so there is only one yml for CI. The problem is that the codecov website will clear the yml files submitted by the previous commit, so only three ymls are executed in one commit to ensure that all three code coverage files are uploaded to the codecov website.

Elssky:
> Now there is a problem: we have three folders: pyspark, spark, cpp need to perform code coverage detection, and add to the three corresponding coverage yml for CI detection and upload to the codecov website, but in most cases every time we Only modify one folder, so there is only one yml for CI. The problem is that the codecov website will clear the yml files submitted by the previous commit, so only three ymls are executed in one commit to ensure that all three code coverage files are uploaded to the codecov website.

I'll try to fix this

PR #603, opened 2024-08-23
fix(dev): add libprotobuf-dev in setup-ubuntu.sh

acezen:
### Reason for this PR

`C++` now rely on protobuf dependencies https://github.com/apache/incubator-graphar/blob/359b82960b368e3780dfda461b8bc984610e8cca/cpp/CMakeLists.txt#L210

add libprotobuf-dev in setup-ubuntu.sh

### Are these changes tested?

yes

no


lixueclaire (approved):
LGTM

PR #610, opened 2024-09-02
input csv and write multi-label chunks

Elssky:


cpp/src/graphar/arrow/chunk_writer.cc
```diff
@@ -251,6 +252,25 @@ Status VertexPropertyWriter::WriteChunk(
...
 }
 
+Status VertexPropertyWriter::WriteLabelChunk(
+    const std::shared_ptr<arrow::Table>& input_table,
+    IdType chunk_index, FileType file_type,
+    ValidateLevel validate_level) const {
+  auto schema = input_table->schema();
+  int indice = schema->GetFieldIndex(GeneralParams::kVertexIndexCol);
+  std::vector<int> indices;
+  for(int i=0; i<schema->num_fields(); i++){
```
acezen, line 262: Is this would select all column?

cpp/src/graphar/arrow/chunk_writer.cc
```diff
@@ -287,9 +307,104 @@ Status VertexPropertyWriter::WriteTable(
...
+                                chunk_index, file_type, validate_level));
+  }
+  return Status::OK();
+}
+
+Result<std::shared_ptr<arrow::Table>> VertexPropertyWriter::GetLabelTable(
+    const std::shared_ptr<arrow::Table>& input_table, 
+    const std::vector<std::string>& labels) const {
+          // Find the :LABEL column index
+    auto label_col_idx = input_table->schema()->GetFieldIndex(":LABEL");
```
acezen, line 355: I suggest put the ":LABEL" to `GeneralParams`.

cpp/src/graphar/arrow/chunk_writer.cc
Elssky: yes, as the input is label chunks, but here exists a bug: we do not need line 260

PR #616, opened 2024-09-03
[WIP] GraphAr Cli

jasinliu:

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,33 @@
+# GraphAr Cli (Under Development)
+
+GraphAr Cli uses [pybind11][] and [scikit-build-core][] to bind C++ code into Python and build command line tools through Python. Command line tools developed using [typer][].
+
+[pybind11]: https://pybind11.readthedocs.io
+[scikit-build-core]: https://scikit-build-core.readthedocs.io
+[typer]: https://typer.tiangolo.com/
+
+## Requirements
+
+- Ubuntu 22.04
+- Cmake >= 3.15
+- Arrow >= 12.0
+- Python >= 3.7
+- pip == latest
+
+
+The best testing environment is `ghcr.io/apache/graphar-dev` Docker environment.
+
+And using Python in conda or venv is a good choice. 
+
+## Installation
+
+- Clone this repository
+- `pip install ./cli` or set verbose level `pip install -v ./cli`
+
+## Usage
+
+```bash
+graphar-cli --help
+```
+
+
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -1,0 +1,56 @@
+[build-system]
+requires = ["scikit-build-core>=0.3.3", "pybind11"]
+build-backend = "scikit_build_core.build"
+
+
+[project]
+name = "graphar_cli"
+version = "0.0.1"
+description = "GraphAr command line tool"
+readme = "README.md"
+authors = [{ name = "GraphAr community", email = "dev@graphar.apache.org" }]
+requires-python = ">=3.7"
+dependencies = ["typer ~= 0.12.3", "pandas ~= 2.2.2", "pydantic ~= 2.8.2"]
+
+
+[project.scripts]
+graphar-cli = "graphar_cli.graphar_cli:main"
+
+
+[tool.scikit-build]
+build-dir = "build"
+
+
+[tool.ruff]
+src = ["src"]
+line-length = 100
+
+
+[tool.ruff.lint]
+extend-select = [
+  "B",   # flake8-bugbear
+  "I",   # isort
+  "ARG", # flake8-unused-arguments
+  "C4",  # flake8-comprehensions
+  "EM",  # flake8-errmsg
+  "ICN", # flake8-import-conventions
+  "G",   # flake8-logging-format
+  "PGH", # pygrep-hooks
+  "PIE", # flake8-pie
+  "PL",  # pylint
+  "PT",  # flake8-pytest-style
+  "PTH", # flake8-use-pathlib
+  "RET", # flake8-return
+  "RUF", # Ruff-specific
+  "SIM", # flake8-simplify
+  "T20", # flake8-print
+  "UP",  # pyupgrade
+  "YTT", # flake8-2020
+  "EXE", # flake8-executable
+  "NPY", # NumPy specific rules
+  "PD",  # pandas-vet
+]
+ignore = [
+  "PLR09",   # Too many X
+  "PLR2004", # Magic comparison
+]
--- a/cli/src/graphar_cli/config.py
+++ b/cli/src/graphar_cli/config.py
@@ -1,0 +1,57 @@
+from typing import Dict, List, Literal, Optional
+
+from pydantic import BaseModel
+
+
+class GraphArConfig(BaseModel):
+    path: str
+    name: str
+    vertexChunkSize: int = 100
+    edgeChunkSize: int = 1024
+    fileType: Literal["parquet", "orc", "csv", "json"] = "parquet"
+    adjListType: Literal[
+        "ordered_by_source", "ordered_by_dest", "unordered_by_source", "unordered_by_dest"
+    ] = "ordered_by_source"
+
+
+class Property(BaseModel):
+    name: str
+    type: Literal["bool", "int32", "int64", "float", "double", "string", "date", "timestamp"]
+    nullable: bool = False
+
+
+class Source(BaseModel):
+    type: Literal["parquet", "orc", "csv", "json"] = "csv"
+    path: str
+    delimiter: str = ","
+    columns: Dict[str, str]
+
+
+class Vertex(BaseModel):
+    type: str
+    labels: List[str]
+    primary: str
+    properties: List[Property]
+    propertyGroups: List[List[str]] = []
+    source: Source
+
+
+class Edge(BaseModel):
+    type: str
+    srcType: str
+    srcProp: str
+    dstType: str
+    dstProp: str
+    properties: List[Property] = []
+    propertyGroups: List[List[str]] = []
+    source: Source
+
+
+class SchemaConfig(BaseModel):
+    vertices: List[Vertex]
+    edges: List[Edge]
+
+
+class ImportConfig(BaseModel):
+    graphar: GraphArConfig
+    import_schema: SchemaConfig
--- a/cli/src/graphar_cli/graphar_cli.py
+++ b/cli/src/graphar_cli/graphar_cli.py
@@ -1,0 +1,157 @@
+import json
+from logging import getLogger
+from pathlib import Path
+from typing import List
+
+import pandas as pd
+import typer
+
+from ._core import (
+    check_edge,
+    check_graph,
+    check_vertex,
+    get_edge_types,
+    get_vertex_types,
+    show_edge,
+    show_graph,
+    show_vertex,
+)
+from .config import ImportConfig
+from .importer import do_import, validate
+from .logging import setup_logging
+
+app = typer.Typer(help="GraphAr Cli", no_args_is_help=True, add_completion=False)
+
+setup_logging()
+logger = getLogger(__name__)
+
+
+@app.command(context_settings={"help_option_names": ["-h", "--help"]}, help="Show the metadata")
+def show(
+    path: str = typer.Option(None, "--path", "-p", help="Path to the GraphAr config file"),
+    vertex: str = typer.Option(None, "--vertex", "-v", help="Vertex type to show"),
+    edge_src: str = typer.Option(None, "--edge-src", "-es", help="Source of the edge type to show"),
+    edge: str = typer.Option(None, "--edge", "-e", help="Edge type to show"),
+    edge_dst: str = typer.Option(
+        None, "--edge-dst", "-ed", help="Destination of the edge type to show"
+    ),
+) -> None:
+    if not Path(path).exists():
+        logger.error("File not found: %s", path)
+        raise typer.Exit(1)
+    path = Path(path).resolve() if Path(path).is_absolute() else Path(Path.cwd(), path).resolve()
+    path = str(path)
+    if vertex:
+        vertex_types = get_vertex_types(path)
+        if vertex not in vertex_types:
+            logger.error("Vertex %s not found in the graph", vertex)
+            raise typer.Exit(1)
+        logger.info(show_vertex(path, vertex))
+        raise typer.Exit(0)
+    if edge or edge_src or edge_dst:
+        if not (edge and edge_src and edge_dst):
+            logger.error("Edge source, edge, and edge destination must all be set")
+            raise typer.Exit(1)
+        edge_types = get_edge_types(path)
+        found = False
+        for edge_type in edge_types:
+            if edge_type[0] == edge_src and edge_type[1] == edge and edge_type[2] == edge_dst:
+                found = True
+                break
+        if not found:
+            logger.error(
+                "Edge type with source %s, edge %s, and destination %s not found in the graph",
+                edge_src,
+                edge,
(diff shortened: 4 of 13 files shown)


cli/README.md
```diff
@@ -0,0 +1,88 @@
...
+The mini version of the configuration file is a simplified version of the full configuration file, retaining the same functionality. It shows the essential parts of the configuration information. 
+
+For the full configuration file, if all fields can be set to their default values, you can simplify it to the mini version. However, it cannot be further reduced beyond the mini version.
+
+
+In the yaml config files, we provide brief comments on the fields, which can be used as a reference
+
+**Example**
+
+To import the movie graph data from the `testing` directory, you first need to prepare data files. Supported file types include `csv`, `json`(as well as`jsonline`, but should have the `.json` extension), `parquet`, and `orc` files. Please ensure the correct file extensions are set in advance, or specify the file type in the source section of the configuration.
```
acezen, line 60: Do we need to the extension to identify the type os source file?

cli/import.full.json
```diff
@@ -0,0 +1,223 @@
+{
```
acezen, line 1: can we put these import configuration files to testing repo? I think the configuration files are better to put together with the data.

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+class AdjList(BaseModel):
+    ordered: bool
+    aligned_by: Literal["src", "dst"]
+    file_type: Literal["parquet", "orc", "csv", "json"]
+
+
+class Edge(BaseModel):
+    edge_type: str
+    src_type: str
+    src_prop: Optional[str] = None
```
acezen, line 165: what are the `src_prop`and `dst_prop` mean here?

cli/src/graphar_cli/graphar_cli.py
```diff
@@ -0,0 +1,187 @@
...
+logger = getLogger(__name__)
+
+
+@app.command(
+    context_settings={"help_option_names": ["-h", "--help"]},
+    help="Show the metadata",
+    no_args_is_help=True,
+)
+def show(
+    path: str = typer.Option(None, "--path", "-f", help="Path to the GraphAr config file"),
```
acezen, line 60: why the full parameter is `--path` but the short is `-f`?

cli/src/main.cc
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <pybind11/pybind11.h>
```
acezen, line 20: use double quotes for third party headers.

cli/src/util.h
```diff
@@ -0,0 +1,483 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <arrow/adapters/orc/adapter.h>
```
acezen, line 22: Ditto.

cli/src/util.h
```diff
@@ -0,0 +1,483 @@
...
+#include <arrow/compute/api.h>
+#include <arrow/csv/api.h>
+#include <arrow/io/api.h>
+#include <arrow/json/api.h>
+#include <graphar/api/arrow_writer.h>
+#include <graphar/api/high_level_writer.h>
+#include <graphar/graph_info.h>
+#include <parquet/arrow/reader.h>
+
+#include <iostream>
```
acezen, line 33: the std library header should put at the begin.

acezen (commented):
you can refer to https://github.com/apache/incubator-graphar/commit/6a13df36224e6c2dcbe70b729ac1cf48498eee7b to fix the ORC support problem in CI.

cli/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+
+
+[project]
+name = "graphar_cli"
+version = "0.0.1"
+description = "GraphAr command line tool"
+readme = "README.md"
+authors = [{ name = "GraphAr community", email = "dev@graphar.apache.org" }]
+requires-python = ">=3.7"
+dependencies = ["typer ~= 0.12.3", "pydantic ~= 2.8.2", "pyyaml ~= 6.0.2"]
```
SemyonSinchenko, line 13: Why did we pin minor versions? As I remember, all these projects are following semantic-versioning and there shouldn't be a problem if we pin only major version.

cli/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+
+[project.optional-dependencies]
+test = ["pandas ~= 2.2.2"]
+
+[project.scripts]
+graphar-cli = "graphar_cli.graphar_cli:main"
```
SemyonSinchenko, line 19: Is a `graphar-cli` a good name? I would expect something like just `graphar`

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from logging import getLogger
+from pathlib import Path
+from typing import Dict, List, Literal, Optional
```
SemyonSinchenko, line 20: The most modern way is to add `from __future__ import annotations` and use `list` and `dict` instead. See [here](https://dev.to/derlin/python-type-hints-and-future-annotations-4kc5#introducing-pep-563-or-what-lies-behind-future-annotations) for details.

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+
+from pydantic import BaseModel, field_validator, model_validator
+from typing_extensions import Self
+
+logger = getLogger("graphar_cli")
+
+# TODO: convert to constants
```
SemyonSinchenko, line 27: I would expect to see these variables in UPPER_CASE

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from logging import getLogger
+from pathlib import Path
+from typing import Dict, List, Literal, Optional
+
+from pydantic import BaseModel, field_validator, model_validator
+from typing_extensions import Self
```
SemyonSinchenko, line 23: For performance reasons all the typing parts should be moved to the `if TYPE_CHEKING:` block. See [here](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) for details.

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+        path = Path(v).resolve().absolute()
+        if not path.exists():
+            path.mkdir(parents=True, exist_ok=True)
+        elif any(path.iterdir()):
+            msg = f"Warning: Path {v} already exists and contains files."
+            logger.warning(msg)
+        return v
+
+
+class Property(BaseModel):
```
SemyonSinchenko, line 59: Why we re-define all this stuff here instead of generating the code from [protobuf](https://github.com/apache/incubator-graphar/blob/main/format/property_group.proto#L29)? How are we going to sync these definitions with protobuf in the future?

cli/src/graphar_cli/logging.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import logging
+from typing import Union
```
SemyonSinchenko, line 19: I thing it should be moved to the `TYPE_CHECKING` block.

cli/src/graphar_cli/importer.py
```diff
@@ -0,0 +1,122 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from logging import getLogger
+
+from .config import ImportConfig
```
SemyonSinchenko, line 20: I thing it can be moved to the `TYPE_CHECKING` block, isn't? I do not see runtime calls to this class

SemyonSinchenko (commented):
Thanks a loot! It looks like a very big step forward for GraphAr. The main question in my mind at the moment is why are we going to re-define basic structures as pyantic models instead of using the code, generated from the protobuf messages? In my understanding the whole idea of proto was to avoid multiple definitions of the same things across the project.

cli/pyproject.toml
Thespica: +1

cli/README.md
jasinliu: The default is to determine the file type based on the file extension. If you do not need to determine the file type based on the file extension, you need to provide the `file_type` field. This option will ignore the file extension.

cli/import.full.json
jasinliu: Sure. I will move it.

cli/src/graphar_cli/config.py
jasinliu: They mean the props in `src_vertex` and `dst_vertex`. We use them to find the `vertex_id`. They should not be `Optional`. I will change them.

cli/pyproject.toml
jasinliu: The `~=` operator will automatically pin the major version number.

cli/src/graphar_cli/config.py
jasinliu: Thanks for sharing! I will make changes in the future.

cli/src/graphar_cli/config.py
jasinliu: Thanks for the great reminder! Now there are some differences between them. I will turn to use it in the next pr

cli/src/graphar_cli/importer.py
jasinliu: Can you provide some examples of using `TYPE_CHECKING`? It would help me a lot, thanks!

SemyonSinchenko: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

cli/src/graphar_cli/config.py
SemyonSinchenko: You can use `buf` tool to generate py-code from protobuf. Feel free to ping me if you need more details or any kind of help with it.

jasinliu: Thanks! The `protobuf` also can be used in `c++` codes in `cli`. It will be a big change.

cli/src/graphar_cli/importer.py
jasinliu: > https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

Thanks! I will try it with the `protobuf` changes.

.github/workflows/cli.yml
```diff
@@ -0,0 +1,116 @@
...
+    
+  #   - name: Build GraphAr And Run Tests
+  #     working-directory: "cli"
+  #     run: |
+  #       pip install ./
+  #       graphar --help
+  #       graphar check -p ../testing/neo4j/MovieGraph.graph.yml
+  #       graphar show -p ../testing/neo4j/MovieGraph.graph.yml -v Person
+  #       graphar show -p ../testing/neo4j/MovieGraph.graph.yml -es Person -e ACTED_IN -ed Movie
+  #       graphar import -c import.full.yml
```
Thespica, line 114: We can disable the job with `if: false`, which is considered better to me, rather than ignore them.

similar to example in [this doc](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow):

image

jasinliu: Great suggestion!  Thanks  very much!

.github/workflows/cli.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    - name: Upload coverage reports to Codecov
+      uses: codecov/codecov-action@v4.0.1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+
+  macos:
+    name: ${{ matrix.architecture }} macOS ${{ matrix.macos-version }} CLI
+    runs-on: macos-${{ matrix.macos-version }}
+    if: false
```
Thespica, line 87: Can you add a note to explain why we skip this job?

Thespica (approved):
LGTM~
Thank you!

acezen:
> Thanks a loot! It looks like a very big step forward for GraphAr. The main question in my mind at the moment is why are we going to re-define basic structures as pyantic models instead of using the code, generated from the protobuf messages? In my understanding the whole idea of proto was to avoid multiple definitions of the same things across the project.

Hi, Sam, The CI currently is based on the C++ library and the protobuf now not full integrate into c++ library yet, so this PR has a little re-define of the base structure but will update when the protobuf is ready.

.github/workflows/cli.yml
acezen: maybe add a TODO

cli/import.full.yml
```diff
@@ -0,0 +1,138 @@
+graphar:                                              # Global configuration of imported data
```
acezen, line 1: the import congfig files should move to testing repo too and it's better to use.the yaml as the only format to specify configurations.

acezen (approved):

acezen (commented):
the import congfig files should move to testing repo too and it's better to use.the yaml as the only format to specify configurations.

cli/src/graphar_cli/config.py
```diff
@@ -0,0 +1,253 @@
...
+DEFAULT_VALIDATE_LEVEL = "weak"
+DEFAULT_VERSION = "gar/v1"
+SUPPORT_FILE_TYPES = {"parquet", "orc", "csv", "json"}
+
+class GraphArConfig(BaseModel):
+    path: str
+    name: str
+    vertex_chunk_size: Optional[int] = 100
+    edge_chunk_size: Optional[int] = 1024
+    file_type: Literal["parquet", "orc", "csv", "json"] = DEFAULT_FILE_TYPE
```
SemyonSinchenko, line 41: I'm just curious what was the reason to use strings instead of creating a enumeration? I the `file_type` in at least four places and it may become challenging to support it in the future... What do you think about something like this:
```python
from enum import Enum

class FileType(str, Enum):
    PARQUET = "parquet"
    ORC = "orc"
    CSV = "csv"
    JSON = "json"
```
and use it instead of the string literals?

SemyonSinchenko (approved):
LGTM overall, I left a minor comment

cli/src/graphar_cli/config.py
jasinliu: `FileType(str, Enum)` with recent versions of `pydantic` will raise a warning. I think it is necessary to pin a minor version.

jasinliu: https://github.com/pydantic/pydantic/discussions/9270

PR #621, opened 2024-09-05
feat(c++): recover the BUILD_ARROW_FROM_SOURCE option

acezen:
### Reason for this PR

There may be senerio that user want to build graphar as a whole static library that include all the dependencies. arrow's static distribution does not include some libraries like thrift, aws...
So we may build arrow from the source.

- recover the `BUILD_ARROW_FROM_SOURCE` to support build arrow from source and link graphar as a single static library that includes all the dependencies.
- add a nightly ci to check the option.

### Are these changes tested?

yes

no

### DIFF
--- a/cpp/cmake/apache-arrow.cmake
+++ b/cpp/cmake/apache-arrow.cmake
@@ -83,6 +83,7 @@ function(build_arrow)
                              "-DARROW_WITH_RE2=OFF"
                              "-DARROW_FILESYSTEM=ON"
                              "-DARROW_CSV=ON"
+                             "-DARROW_JSON=ON"
                              "-DARROW_PYTHON=OFF"
                              "-DARROW_BUILD_BENCHMAKRS=OFF"
                              "-DARROW_BUILD_TESTS=OFF"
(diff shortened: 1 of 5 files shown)


cpp/cmake/apache-arrow.cmake
```diff
@@ -107,11 +108,11 @@ function(build_arrow)
...
-    if (Arrow_FOUND) # arrow is installed, build the same version as the installed one
-        message(STATUS "Found Arrow installed, align to version: ${Arrow_VERSION}")
-        set(ARROW_VERSION_TO_BUILD "${Arrow_VERSION}" CACHE INTERNAL "arrow version")
-    endif ()
+    set(ARROW_VERSION_TO_BUILD "15.0.0" CACHE INTERNAL "arrow version")
+    # if (Arrow_FOUND) # arrow is installed, build the same version as the installed one
+        # message(STATUS "Found Arrow installed, align to version: ${Arrow_VERSION}")
+        # set(ARROW_VERSION_TO_BUILD "${Arrow_VERSION}" CACHE INTERNAL "arrow version")
+    # endif ()
     set(GAR_ARROW_SOURCE_FILE "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_VERSION_TO_BUILD}/apache-arrow-${ARROW_VERSION_TO_BUILD}.tar.gz")
```
jasinliu, line 116: Can environment variables or parameters be passed here?

jasinliu: If the user has their own download link, they can identify and determine the version after installation.

> Can environment variables or parameters be passed here?

acezen: hi, @jasinliu, thanks for the advice. Maybe we can not expose the variable to user, and if there is arrow installed in the system, we can align to the version

acezen: > If the user has their own download link, they can identify and determine the version after installation.
> 
> > Can environment variables or parameters be passed here?

and I not sure what you mean own download link? you mean there maybe arrow that modified by user?

jasinliu: > > If the user has their own download link, they can identify and determine the version after installation.
> > > Can environment variables or parameters be passed here?
> 
> and I not sure what you mean own download link? you mean there maybe arrow that modified by user?

For example, the specified mirror link， the modification of the arrow，or some test version of the arrow. Custom links can provide some flexibility。

acezen: > > > If the user has their own download link, they can identify and determine the version after installation.
> > > > Can environment variables or parameters be passed here?
> > 
> > 
> > and I not sure what you mean own download link? you mean there maybe arrow that modified by user?
> 
> For example, the specified mirror link， the modification of the arrow，or some test version of the arrow. Custom links can provide some flexibility。

But we are supposed to only support the original version of arrow, provide a link for that is a little weird. I suggest we still use the link from apache, and if user want to use other link, he could modify the cmake file directly.

lixueclaire (approved):
LGTM

PR #625, opened 2024-09-10
chore(c++): Support ENV GAR_ARROW_SOURCE_URL to let user set arrow source url

acezen:

### DIFF
--- a/cpp/cmake/apache-arrow.cmake
+++ b/cpp/cmake/apache-arrow.cmake
@@ -110,14 +110,17 @@ function(build_arrow)
     find_package(Arrow QUIET)
     set(ARROW_VERSION_TO_BUILD "15.0.0" CACHE INTERNAL "arrow version")
     if (Arrow_FOUND) # arrow is installed, build the same version as the installed one
-       message(STATUS "Found Arrow installed, align to version: ${Arrow_VERSION}")
        set(ARROW_VERSION_TO_BUILD "${Arrow_VERSION}" CACHE INTERNAL "arrow version")
     endif ()
-    set(GAR_ARROW_SOURCE_FILE "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_VERSION_TO_BUILD}/apache-arrow-${ARROW_VERSION_TO_BUILD}.tar.gz")
+    if(DEFINED ENV{GAR_ARROW_SOURCE_URL})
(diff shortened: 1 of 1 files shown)


cpp/cmake/apache-arrow.cmake
```diff
@@ -108,16 +108,19 @@ function(build_arrow)
...
-    set(GAR_ARROW_SOURCE_FILE "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_VERSION_TO_BUILD}/apache-arrow-${ARROW_VERSION_TO_BUILD}.tar.gz")
+    set(ARROW_VERSION_TO_BUILD "13.0.0" CACHE INTERNAL "arrow version")
+    # if (Arrow_FOUND) # arrow is installed, build the same version as the installed one
+    #    set(ARROW_VERSION_TO_BUILD "${Arrow_VERSION}" CACHE INTERNAL "arrow version")
+    # endif ()
+    if(DEFINED ENV{GAR_ARROW_SOURCE_URL})
+        set(GAR_ARROW_SOURCE_URL "$ENV{GAR_ARROW_SOURCE_URL}")
+    else()
+        set(GAR_ARROW_SOURCE_URL "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_VERSION_TO_BUILD}/apache-arrow-${ARROW_VERSION_TO_BUILD}.tar.gz")
+    endif()
```
jasinliu, line 119:
```suggestion
    if(DEFINED ENV{GAR_ARROW_SOURCE_URL})
        set(GAR_ARROW_SOURCE_URL "$ENV{GAR_ARROW_SOURCE_URL}")
    else()
         set(ARROW_VERSION_TO_BUILD "13.0.0" CACHE INTERNAL "arrow version")
    # if (Arrow_FOUND) # arrow is installed, build the same version as the installed one
    #    set(ARROW_VERSION_TO_BUILD "${Arrow_VERSION}" CACHE INTERNAL "arrow version")
    # endif ()
        set(GAR_ARROW_SOURCE_URL "https://www.apache.org/dyn/closer.lua?action=download&filename=arrow/arrow-${ARROW_VERSION_TO_BUILD}/apache-arrow-${ARROW_VERSION_TO_BUILD}.tar.gz")
    endif()
```

jasinliu: If arrow url is provided by environment variable, we should not specify arrow version

jasinliu (approved):

PR #650, opened 2024-10-28
feat(c++): add function and example for read and write label chunk

Elssky:
### Reason for this PR

add function for support label related fuction.

chunk_reader, chunk_writer and test cases.
### Are these changes tested?

yes


cpp/src/graphar/general_params.h
Elssky, line 1: After considering, here I changed the parameter name back to ': LABEL', mainly because it is consistent with the column name in the original csv file, so as to avoid confusing users when operating arrow tables.

Elssky:
@lixueclaire please review :)

lixueclaire (approved):
LGTM

Issue #662, opened 2025-03-01
feat(java): shade protobuf dependency in Java info


SemyonSinchenko:
### Describe the enhancement requested

Protobuf is very popular but version-sensitive dependency that is used literally everywhere. We should shade it in Java info to simplify the integration of the GAR into other projects. For example, we can simply rename all the `com.google.protobuf.*` to something like `org.grapharproject.protobuf.*`. It may be easily done via Maven assembly plugin.

### Component(s)

Java

SemyonSinchenko:
@Thespica @acezen what do you think?

Thespica:
> [@Thespica](https://github.com/Thespica) [@acezen](https://github.com/acezen) what do you think?

Is it means that we generate the proto-code once and save them, so we no longer need protobuf related dependencies when releasing?

SemyonSinchenko:
> Is it means that we generate the proto-code once and save them, so we no longer need protobuf related dependencies when releasing?

Yes. All the needed classes will be packed into the assembly and renamed by pattern. But it will happens after the compilation and we are still needing the protobuf-maven and protoc to generate the code during the release.

Release:
1. Code generation by protoc
2. Compilation
3. Assembly + renaming

SemyonSinchenko:
I can actually open a draft PR to explain it by the code

SemyonSinchenko:
@Thespica You can check in #663

SemyonSinchenko:
Exactly! I slightly changed the pom (will describe in the future commit). And I have now two generated JARs:
- info
- original info

![Image](https://github.com/user-attachments/assets/d79b49a9-1e9c-433c-93cd-48539d9742d3)

Info contains all the proto-related classes:

![Image](https://github.com/user-attachments/assets/aa67ddb8-fbcd-4d15-95b2-48e6823266d1)

I can pick one and decompile to see that everything was renamed in a right way:
![Image](https://github.com/user-attachments/assets/40fb84f2-1c93-47ad-b3e5-8da4439f9134)

So, with such an approach we do not need to have a separate `proto` subproject and at the same time we allows our user to do integrations without dependencies conflicts.

SemyonSinchenko:
@Thespica cc

Thespica:
> [@Thespica](https://github.com/Thespica) cc

Is this PR ready? 
It seems that the `dependency-reduce-pom.xml` file is temporary, and we do not need to add it to git(from https://stackoverflow.com/a/35155294)

PR #663, opened 2025-03-01
feat: shade proto

SemyonSinchenko:
### Reason for this PR

As described in #662 

Only shading rules

### Are these changes tested?

Changes are related to pom.xml only

No


Thespica:
Looks cool!
By shading like this, we can get a jar with info package, right?

SemyonSinchenko:
> Looks cool! By shading like this, we can get a jar with info package, right?

I answered in the issue

Thespica:
Is this PR ready?
It seems that the dependency-reduce-pom.xml file is temporary, and we do not need to add it to git(from [stackoverflow.com/a/35155294](https://stackoverflow.com/a/35155294))

SemyonSinchenko:
@Thespica Yes! From now it is ready for the review.

**Changes:**
- Shading snakeyaml too (it is a hadoop dependency and may cause dependency hell)
- Updating .gitignore
- Drop tmp pom
- Typos

Issue #684, opened 2025-05-23
bug(java, info): Parameter order mismatch between EdgeInfo constructor and EdgeYaml#toEdgeInfo


yangxk1:
### Describe the bug, including details regarding any error messages, version, and platform.

The third issue discussed in #676 

In the `EdgeInfo` constructor, the parameter `srcChunkSize` appears before `chunkSize`. However, in `org.apache.graphar.info.yaml.EdgeYaml#toEdgeInfo`, `srcChunkSize` is passed after `chunkSize`.

### Component(s)

Java

Issue #676, opened 2025-04-11
feat(java, info): Unit test for loading graph


Thespica:
### Describe the enhancement requested

Write unit test for java-info module, loading graph from [test set](https://github.com/apache/incubator-graphar-testing/tree/955596c325ceba7b607e285738e3dd0ce4ff424e), ensuring that info of:
- [ ] GraphInfo / VertexInfo / EdgeInfo
- [ ] PropertyGroup
- [ ] AdjacentList
- [ ] Property

are loaded as expected.

Can reference [info test of C++](https://github.com/apache/incubator-graphar/blob/main/cpp/test/test_info.cc) and [info test of scala](https://github.com/apache/incubator-graphar/blob/main/maven-projects/spark/graphar/src/test/scala/org/apache/graphar/TestGraphInfo.scala).

### Component(s)

Java

yangxk1:
I noticed that some interfaces in the Java code are not fully aligned with those in the C++ code. For example, the [Java code](https://github.com/apache/incubator-graphar/blob/main/maven-projects/java/src/main/java/org/apache/graphar/graphinfo/VertexInfo.java) is missing the `vertexInfo::getPropertyGroupByIndex` interface, which is available in the [C++ code](https://github.com/apache/incubator-graphar/blob/main/cpp/src/graphar/graph_info.h) . Also, the current Java project depends on version `v0.10.0` of the C++ code, which is quite outdated. Should we consider updating the Java project first?

Thespica:
> I noticed that some interfaces in the Java code are not fully aligned with those in the C++ code. For example, the [Java code](https://github.com/apache/incubator-graphar/blob/main/maven-projects/java/src/main/java/org/apache/graphar/graphinfo/VertexInfo.java) is missing the `vertexInfo::getPropertyGroupByIndex` interface, which is available in the [C++ code](https://github.com/apache/incubator-graphar/blob/main/cpp/src/graphar/graph_info.h) . Also, the current Java project depends on version `v0.10.0` of the C++ code, which is quite outdated. Should we consider updating the Java project first?

Hi @yangxk1 , maybe there are some ambiguous expression. The `java-info` here refer to the [info](https://github.com/apache/incubator-graphar/tree/main/maven-projects/info) packages, rather than the [original java](https://github.com/apache/incubator-graphar/tree/main/maven-projects/java)  package. You can reach out to us if there is any other problem : )

yangxk1:
[...]

PR #685, opened 2025-05-23
fix(java,info): Reorder parameters in EdgeInfo constructor

yangxk1:
## Reason for this PR

According to the description in #684 , parameter order mismatch between EdgeInfo constructor and EdgeYaml#toEdgeInfo.

## What changes are included in this PR?

Reordered EdgeInfo constructor parameters to match C++ version

## Are these changes tested?

yes

## Are there any user-facing changes?

no


yangxk1:
Hi, @Thespica. This should solve the problem in #684 #676 , please review this.

Thespica:
Hi @yangxk1 , I think order of `srcChunkSize` -> `chunkSize` -> `dstChunkSize` will be better.

yangxk1:
Thanks for your suggestion @Thespica ! But looking at the C++ constructor for [EdgeInfo](https://github.com/apache/incubator-graphar/blob/main/cpp/src/graphar/graph_info.cc), the order is `edgeChunkSize` → `srcChunkSize` → `dstChunkSize`. I think keeping the same order in Java would be more intuitive and user-friendly. 
edge_info

yecol (approved):

Issue #688, opened 2025-05-23
fix(java): adjust children module version of pom to avoid warning


Thespica:
### Describe the enhancement requested

When running maven-projects, there are warning like:
```bash
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.apache.graphar:graphar-info:jar:0.12.0-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ org.apache.graphar:graphar-info:${graphar.version}, /home/john/graphar/incubator-graphar/maven-projects/info/pom.xml, line 37, column 14
[WARNING] 'version' contains an expression but should be a constant. @ org.apache.graphar:graphar-root:${graphar.version}, /home/john/graphar/incubator-graphar/maven-projects/pom.xml, line 32, column 14
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

should remove version property, add add comment to notice that.

refer: https://stackoverflow.com/questions/1981151/warning-on-using-project-parent-version-as-the-version-of-a-module-in-maven-3

### Component(s)

Java, Spark

PR #689, opened 2025-05-26
fix(java): adjust children module version of pom to avoid warning

sanyakapoor27:
### Reason for this PR
 As described in #688, this PR is done to remove the maven warnings due to version property which may break future maven builds.

Removed the version property from child modules where it can be inherited from the parent and replaced expressions with constants in <parent><version>.

### Are these changes tested?
 yes, verified maven build after the changes.

no


yecol:
Hi @sanyakapoor27, thanks for your commit!

However, I suggest not adding version constants in all child POMs, as this could make maintenance more difficult. It’s better to remove the version tags from child POMs.

yecol (changes requested):
See comments.

sanyakapoor27:
@yecol thanks for the review once again! I had previously updated line 32 of the root POM to <version>0.12.0-SNAPSHOT</version>, but that led to "[FATAL] Non-resolvable parent POM" errors when building the child modules. 

This error and the warning mentioned in the issue #688 were gone when the <version> in the child poms also referred to 0.12.0-SNAPSHOT. This can be viewed here: [Stack Overflow](https://stackoverflow.com/questions/13500332/non-resolvable-parent-pom-could-not-transfer-artifact-when-trying-to-refer-to). In such a case, please let me know what would be the best course of action?

keksmd (commented):
@sanyakapoor27 I reproduced this locally end to end, so here is the full picture including an answer to your open question about `Non-resolvable parent POM`.

## The current diff does not remove the warning

The warning is reported against the **root** POM's own `<version>`, not against the children:

```
[WARNING] 'version' contains an expression but should be a constant.
    @ org.apache.graphar:graphar-root:${graphar.version}, maven-projects/pom.xml, line 32, column 14
```

This PR only deletes `<version>${graphar.version}</version>` from the child modules and leaves `maven-projects/pom.xml` line 32 untouched, so the warning survives. Removing the redundant child `<version>` elements is still the right thing to do (@yecol's point about maintenance) — it just isn't sufficient on its own.

## Why hardcoding line 32 alone gave you `Non-resolvable parent POM`

Maven resolves the `<parent>` coordinates *before* it can interpolate any properties, because properties live in the parent it hasn't loaded yet. So `<parent><version>` must be a literal. Every child in this repo currently has:

```xml
<parent>
    <groupId>org.apache.graphar</groupId>
    <artifactId>graphar-root</artifactId>
    <version>${graphar.version}</version>
    <relativePath>../pom.xml</relativePath>
</parent>
```

I reproduced your exact failure by hardcoding only line 32:

```
[FATAL] Non-resolvable parent POM for org.apache.graphar:graphar-info:${graphar.version}:
        The following artifacts could not be resolved:
        org.apache.graphar:graphar-root:pom:${graphar.version} (absent)
```

Note the coordinate: `graphar-root:pom:${graphar.version}` — the literal string is being used as the version. It is not a repository or connectivity problem; the parent version expression simply never expands.

## The fix that works

Replace **every** `${graphar.version}` occurrence that sits in a `<version>` element with the literal, in all of:

- `maven-projects/pom.xml` (its own `<version>`, line 32)
- `maven-projects/{info,java,storage-api,spark}/pom.xml` (the `<parent><version>`)
- `maven-projects/spark/{datasources-32,datasources-33,datasources-34,datasources-35,graphar}/pom.xml` (the `<parent><version>`)

and drop the now-redundant child `<version>` elements you already removed in this PR (children inherit the parent version).

I applied that to current `main` and the whole reactor validates cleanly:

```
[INFO] Reactor Build Order:
[INFO] Building Apache GraphAr Root POM 0.13.0-SNAPSHOT   [1/7]
...
[INFO] BUILD SUCCESS
```

with no `'version' contains an expression` warning and no `Non-resolvable parent POM`.

The `graphar.version` property itself can stay defined — it is still useful for `<dependency>` versions (for example `maven-projects/spark/graphar/pom.xml` line 42), where property interpolation works fine. Only `<version>` and `<parent><version>` elements need to be literals.

## Two notes on rebasing

1. This PR was opened against an older `main`. Since then `maven-projects/storage-api` has been added and it carries the same `${graphar.version}` pattern in both places, so it needs the same treatment.
2. There is a second, unrelated warning in the reactor that this PR should *not* try to absorb:

```
[WARNING] 'build.pluginManagement.plugins.plugin.(groupId:artifactId)' must be unique
    but found duplicate declaration of plugin org.apache.maven.plugins:maven-install-plugin
    @ line 257, column 25
```

Worth its own issue.

Issue #397, opened 2024-03-10
[Feat][C++] Support read certain set of properties in the file level, not with `arrow::Table::Select`


acezen:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

yangxk1:
Hi, @acezen ! 
I think we can improve this by setting `ScannerBuilder.Project`. Compared to filtering at the `FileReader` level, although it may have some impact on performance, it will not affect the existing functionality. 
I have conducted a simple test confirming the optimization effect when `ScannerBuilder.Project` is used.
I will open a PR to solve it.

PR #697, opened 2025-05-29
[WIP] feat(c++): support reading certain set of properties

yangxk1:
### Reason for this PR

This PR addresses the feature request in #397  to support reading certain set of properties.

I have modified `VertexPropertyArrowChunkReader::Make` so that when a single `property_name` is provided, it no longer loads all properties in the corresponding `propertyGroup`, but only reads the **internal ID** and the specified **property column**.
Additionally, a `std::vector<std::string>` of `property_names` can be passed to read a specific set of properties, provided they all belong to the same `propertyGroup`.

Essentially, these `property_names` are added to `FilterOptions.columns`, serving as conditions for predicate pushdown.

### Are these changes tested?

Yes, I have added:

- Examples
- Unit tests
- Benchmarks

The benchmark results show performance improvements, especially on large chunk sizes.

- `ReadChunkSelectAllColumnsIn*`: read using a property group (3 properties + internal ID) 
- `ReadChunkSelectOneColumnIn*`: read only the internal ID and a single property
- `ReadChunkSelectTwoColumnIn*`: read only the internal ID and 2 properties

- `*FirstGraph`: read person vertex in ldbc_sample (chunk_size: 100)
- `*SecondGraph`: read organisation vertex in ldbc (chunk_size: 4096)

image

Yes, there are breaking changes:

Previously, users could pass a `property_name` directly to access data from its `propertyGroup`.
Now, to read `propertyGroup`, users should explicitly provide the corresponding `propertyGroup`.
This change improves clarity and ensures consistency when filtering properties.


yangxk1:
In addition, I’ve written a demo showing how to use the `reader` to read specific properties. The implementation can be found at: https://github.com/yangxk1/incubator-graphar/tree/test-select-columns-by-reader

The demo demonstrates that the `reader` offers significant performance gains compared to the current `scanner`-based approach. 
However, the reader does not support predicate pushdown.

`*V2` use the `reader`
image

PR #713, opened 2025-06-27
feat(Java,info): VertexInfo/EdgeInfo Version Info

sapienza88:
This PR introduces parsing of vertex version info

### DIFF
--- a/maven-projects/info/src/main/java/org/apache/graphar/info/GraphInfo.java
+++ b/maven-projects/info/src/main/java/org/apache/graphar/info/GraphInfo.java
@@ -85,7 +85,7 @@ public class GraphInfo {
     }
 
     public String dump() {
-        Yaml yaml = new Yaml(GraphYaml.getDumperOptions());
+        Yaml yaml = new Yaml(GraphYaml.getRepresenter(), GraphYaml.getDumperOptions());
         GraphYaml graphYaml = new GraphYaml(this);
         return yaml.dump(graphYaml);
     }
@@ -189,9 +189,6 @@ public class GraphInfo {
     }
 
     public VersionInfo getVersion() {
-        System.out.println("====================");
-        System.out.println(protoGraphInfo.getVersion());
-        System.out.println("====================");
         return VersionParser.getVersion(protoGraphInfo.getVersion());
     }
 
--- a/maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
+++ b/maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
@@ -208,7 +208,7 @@ public class EdgeInfo {
     }
 
     public String dump() {
-        Yaml yaml = new Yaml(GraphYaml.getDumperOptions());
+        Yaml yaml = new Yaml(GraphYaml.getRepresenter(), GraphYaml.getDumperOptions());
         EdgeYaml edgeYaml = new EdgeYaml(this);
         return yaml.dump(edgeYaml);
     }
--- a/maven-projects/info/src/main/java/org/apache/graphar/info/VersionInfo.java
+++ b/maven-projects/info/src/main/java/org/apache/graphar/info/VersionInfo.java
@@ -20,21 +20,53 @@
 package org.apache.graphar.info;
 
 import java.util.List;
+import java.util.Map;
+import java.util.StringJoiner;
 
 public class VersionInfo {
-    public int parsedVersion;
-    public List<String> parsedUserDefinedTypes;
+    private int version;
+    private List<String> userDefinedTypes;
+    private final Map<Integer, List<String>> version2types =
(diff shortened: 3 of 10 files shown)


yangxk1:
Thanks for the PR — the `version` support you added is really useful!

I just wanted to clarify one thing:
I think my earlier explanation wasn't very clear. "property filtering", I meant filtering properties during data reading using predicate pushdown, e.g., reading only certain properties from Parquet/CSV/ORC files.

some small suggestions:
- Could `getVersion()` return `InfoVersion` directly?
- I think it would make sense to move `InfoVersion` out of `VersionParser`, as it's part of the public API, but `VersionParser` is more of a tool class.

unical1988:
> * getVersion()

I made a static method getVersion() for getting the InfoVersion from the version string as read from Yaml.
But do we need to read the data files line by line? how do you read data files given their attributes (columns) are specified in a Yaml model file? for example given https://github.com/apache/incubator-graphar-testing/blob/main/ldbc_sample/csv/person.vertex.yml how to create an object where the attributes are those specified in .yml as to load the data per chunk?

unical1988:
> Thanks for the PR — the `version` support you added is really useful!
> 
> I just wanted to clarify one thing: I think my earlier explanation wasn't very clear. "property filtering", I meant filtering properties during data reading using predicate pushdown, e.g., reading only certain properties from Parquet/CSV/ORC files.
> 
> some small suggestions:
> 
> * Could `getVersion()` return `InfoVersion` directly?
> * I think it would make sense to move `InfoVersion` out of `VersionParser`, as it's part of the public API, but `VersionParser` is more of a tool class.

getVersion() are talking about graph version or vertex version ?

maven-projects/info/src/main/java/org/apache/graphar/info/PropertyGroup.java
```diff
@@ -221,6 +222,24 @@ boolean isPrimaryKey(String propertyName) {
         return properties.get(propertyName).isPrimary();
     }
 
+    Map<String, Property> getPropertiesByDataType(DataType type) {
```
yangxk1, line 225: These methods can be deleted to keep the PR focused on a single functionality.

yangxk1 (changes requested):

yangxk1:
> > * getVersion()
> 
> I made a static method getVersion() for getting the InfoVersion from the version string as read from Yaml. But do we need to read the data files line by line? how do you read data files given their attributes (columns) are specified in a Yaml model file? for example given https://github.com/apache/incubator-graphar-testing/blob/main/ldbc_sample/csv/person.vertex.yml how to create an object where the attributes are those specified in .yml as to load the data per chunk?

Please refer to [LocalYamlGraphLoader](https://github.com/apache/incubator-graphar/blob/main/maven-projects/info/src/main/java/org/apache/graphar/info/loader/LocalYamlGraphLoader.java#L43).

unical1988:
> > @yangxk1 have you had the chance to look at my PR and review them?
> 
> The CI checks is failing https://github.com/apache/incubator-graphar/pull/713/checks

I know i will take care of that, but the reviews about the code not the CI

yangxk1:
The CI failure might be due to syntax errors, and the code is still subject to significant changes.
Therefore, conducting a detailed code review before the CI passes may not be meaningful.

unical1988:
> The CI failure might be due to syntax errors, and the code is still subject to significant changes. Therefore, conducting a detailed code review before the CI passes may not be meaningful.

No it is not due to syntax error. Pls the rules of Apache say differently

yangxk1 (approved):

unical1988:
@yangxk1 thank you for approving this, I will check the CI and license error for this to be merged.

unical1988:
@yangxk1 I will re-check the functions used in other PRs (I didn't use AI code).

unical1988:
@yangxk1 any idea why `protoGraphInfo.getVersion()` does not return the graph version in `GraphInfoTest.java`?

maven-projects/info/src/main/java/org/apache/graphar/info/GraphInfo.java
```diff
@@ -183,6 +183,12 @@ public String getPrefix() {
         return protoGraphInfo.getPrefix();
     }
 
+    // TODO test protoGraphInfo why is not returning the version from the graph yaml file?
+    public VersionInfo getVersion() {
+        return VersionParser.getVersion(
```
yangxk1, line 188: should not set `gar/v1` as the default version. In c++ library if version does not exist in yaml, the version will be set to `nullptr` (as NULL in JAVA)

yangxk1:
If you open `Allow edits by maintainers` , I can help you to debug

yangxk1:
ok, pls wait a moment

yangxk1:
I will merge this pr, thanks for your contribution! @unical1988

Issue #733, opened 2025-08-18
bug(c++): EdgeIter causes property misalignment, IO errors, and crashes when iterating across chunks or using segmented jumps


gary-cloud:
### Describe the bug, including details regarding any error messages, version, and platform.

When reading edges from `graphar::EdgesCollection` / `EdgeIter` in **segmented jump** or **batch scan** mode, edge properties such as `creationDate` become **misaligned** (out of sync with the order seen in `high_level_example`). In some cases, `EdgeIter` attempts to open non-existent chunk files, leading to `runtime_error` / `abort` (e.g., `Failed to open ... part10/chunk0`) or AddressSanitizer errors.

In contrast, the `high_level_example` sequential one-pass iteration (`auto edge = *it`) usually appears correct, but repeated segmented traversal or direct calls to `it.property()` trigger the bug.

---

### Steps to Reproduce (Minimal Example)

Using GraphAr sample data (`ldbc_sample`), the issue can be reproduced with the following pseudocode:

```cpp
auto edges = graphar::EdgesCollection::Make(..., graphar::AdjListType::ordered_by_source).value();
auto begin = edges->begin();
auto end = edges->end();

size_t count = 0;
for (auto it = begin; it != end; ++it) {
    if (count > 2000) continue;
    count++;
    std::cout << it.source() << "," << it.destination()
              << "," << it.property<std::string>("creationDate").value() << std::endl;
}

// Register a new EdgeIter and perform segmented traversal
auto begin2 = edges->begin();
int i = 0;
for (auto it = begin2; it != end; ++it, i++) {
  if (i <= 2000) continue;
  if (i > 4000) break;

  count++;
  std::cout << "src=" << it.source() << ", dst=" << it.destination()
            << ", creationDate="
            << it.property<std::string>("creationDate").value()
            << std::endl;
}
```

On the second or later segmented traversal, `creationDate` values become misaligned. When crossing vertex-chunk boundaries, the program may attempt to open a non-existent file and crash with `IOError: failed to open local file ...`.

---

### Expected Behavior

* Properties read via **any traversal mode** (`auto edge = *it`, `it.property(...)`, segmented/jump iteration, batch/parallel scans) should always match the results of a **single sequential traversal**.
* Iteration should never attempt to open non-existent chunk files, nor crash.

---

### Actual Behavior

* Property values become misaligned starting from a certain position.
* Crossing chunk/vertex-chunk boundaries may trigger attempts to open invalid file paths, causing crashes.

---

[...]

PR #736, opened 2025-08-24
fix(c++): keep property readers in sync when crossing full edge-chunk boundaries in EdgeIter::operator++

gary-cloud:
### Reason for this PR
When `EdgeIter` advances to the next edge chunk (exactly when `cur_offset_ % chunk_size_ == 0`), property readers were not advanced in that code path. The missing advancement could cause `adj_list_reader_` and `property_readers_` to become unsynchronized, which then produced wrong property values or file-not-found errors. This PR ensures property readers are advanced in that boundary case. (relevant issue: #733) 
##### Minimal Reproducible Example (for dataset `ldbc_sample`):
```c++
auto begin = edges->begin();
int i = 0;
for (auto it = begin; it != end; ++it, i++) {
  if (i <= 4000) {
    continue;
  }
  if (i > 6000) {
    break;
  }
  count++;
  std::cout << "src=" << it.source() << ", dst=" << it.destination()
            << ", creationDate="
            << it.property<std::string>("creationDate").value()
            << std::endl;
}
```

When the iterator advances into a new edge-chunk (normal path, not the KeyError/overflow path), call `reader.next_chunk()` for each `AdjListPropertyArrowChunkReader` so property readers remain aligned with the adjacency reader. **The occurrence of this bug is not attributable to shared state among multiple edge iterators.**

### Are these changes tested?
Yes

No

### DIFF
--- a/cpp/src/graphar/filesystem.cc
+++ b/cpp/src/graphar/filesystem.cc
@@ -211,8 +211,8 @@ Result<T> FileSystem::ReadFileToValue(const std::string& path) const noexcept {
 }
 
 template <>
-Result<std::string> FileSystem::ReadFileToValue(
-    const std::string& path) const noexcept {
+Result<std::string> FileSystem::ReadFileToValue(const std::string& path) const
+    noexcept {
   GAR_RETURN_ON_ARROW_ERROR_AND_ASSIGN(auto access_file,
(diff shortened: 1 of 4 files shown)


yangxk1:
LGTM! Thank you~

Issue #740, opened 2025-09-02
feat(java, info): builder design pattern for java-info


Thespica:
### Describe the enhancement requested

Naive constructors are not friendly for users when the function need such more parameters

https://github.com/apache/incubator-graphar/blob/09703b6fc3ed3499fc8ed1b410647e50c96ec506/maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java#L54-L65

The builder design pattern will help.

### Component(s)

Java

Jarvx200:
Hey, could you assign me for this issue?

Jarvx200:
Hey there, I submitted a pr of the proposed structure to keep the API access level consistent.
It that's ok I will try looking to abstract the logic of the builder and specialize implementations to reduce boilerplate.

PR #743, opened 2025-09-03
feat(java,info): refactor `EdgeInfo` creation using access-level compliant builder, a…

Jarvx200:
### Reason for this PR

Boiler plate heavy telescoping constructor inside `EdgeInfo` #740 

Refactor of telescoping constructor to builder pattern while keeping API access levels and compatibility.

### Are these changes tested?

Yes, using the new builder in `TestUtil` and running the dependent `GraphSaverTest`

No

### DIFF
--- a/maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
+++ b/maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
@@ -19,6 +19,15 @@
 
 package org.apache.graphar.info;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.graphar.info.type.AdjListType;
 import org.apache.graphar.info.type.DataType;
 import org.apache.graphar.info.yaml.EdgeYaml;
@@ -32,15 +41,6 @@ import org.apache.hadoop.fs.Path;
 import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class EdgeInfo {
     private final EdgeTriplet edgeTriplet;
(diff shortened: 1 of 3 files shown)


Jarvx200:
If the pattern is correct, I can continue by deprecating the old constructors and switching to the new method

yangxk1 (commented):
To be honest, [`Lombok`](https://projectlombok.org/) is very suitable for adding builder, but it will bring more dependencies

Jarvx200:
Ok sure, I will start working on the first two problems, the third one would imply temporary builder fields that will eventually get transformed with null checking to triplet, but I can wait for further insights on the problem and can implement if ok.

Jarvx200:
Changes:
- Added static factory to builder as requested 
- Added methods for list types to get appended (Note: since `PropertyGroups` is immutable reducing was needed to get the last instance with all the appended groups)
- Added support to construct `EdgeTriplet` with variables inside builder

yangxk1 (commented):
The overall code is good, I proposed some small issues, and it will be better after solving!

Jarvx200:
Hello, I went through all the proposed issues and provided fixes and testing!

maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -51,6 +53,192 @@ public class EdgeInfo {
...
+        }
+
+        public EdgeInfoBuilder addPropertyGroups(List<PropertyGroup> propertyGroups) {
+            if(propertyGroupsAsListTemp == null)
+                propertyGroupsAsListTemp = new ArrayList<>();
+            propertyGroupsAsListTemp.addAll(propertyGroups);
+            return this;
+        }
+
+        private EdgeInfoBuilder addPropertyGroups(PropertyGroups propertyGroups) {
```
yangxk1, line 167: change the access modifiers and name to `public propertyGroups`

maven-projects/info/src/main/java/org/apache/graphar/info/EdgeInfo.java
```diff
@@ -51,6 +53,192 @@ public class EdgeInfo {
...
+
+        public EdgeInfoBuilder adjacentLists(List<AdjacentList> adjacentListsAsList) {
+            if(adjacentListsAsListTemp == null) {
+                adjacentListsAsListTemp = new ArrayList<>();
+            }
+            this.adjacentListsAsListTemp.addAll(adjacentListsAsList);
+            return this;
+        }
+
+        private EdgeInfoBuilder adjacentLists(Map<AdjListType, AdjacentList> adjacentLists) {
```
yangxk1, line 148: Also public

yangxk1 (commented):
Hi @Jarvx200, Your code is already very good, but needs to format and add license to pass ci test. And I found that there is some  mistake about access modifiers and name. 
It would be better if you could add more tests (e.g., for addPropertyGroup` and `addAdjacentlists` )

Jarvx200:
Hello! I will start working on those problems, write some more tests to make sure everything is compliant.

Jarvx200:
Hello, I fixed the issues above and added a missing `final` to `TestUtil`. If there is any further testing required please tell me.

PR #745, opened 2025-09-04
[WIP] refactor GraphInfoLoader

yangxk1:


yangxk1:
Hi @SemyonSinchenko, Please help me see if the implementation of GraphInfoLoader is correct (ReaderGraphInfoLoader, StreamGraphInfoLoader, StringGraphInfoLoader).
I added an example impl in the tests: maven-projects/info/src/test/java/org/apache/graphar/info/LocalFileSystemStringStreamLoader.java
thanks!

SemyonSinchenko (commented):
I think we really may make it much more simpler by delegate resolving to `java.net.URI`

maven-projects/info/src/main/java/org/apache/graphar/util/PathUtil.java
```diff
@@ -20,8 +20,22 @@
 package org.apache.graphar.util;
 
 public class PathUtil {
+
+    public static String resolvePath(String basePath, String subPath) {
```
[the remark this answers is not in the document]
yangxk1: Hi @SemyonSinchenko , Do we use URI in full scenarios or just use URI to resolve the path.
If it is the former, does the chunk path also apply to uri (for example, `getPropertyGroupChunkPath`). If it is the latter, `Path.of(basePath).resolve(subPath).toString()` can also be easily solved.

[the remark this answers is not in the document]
SemyonSinchenko: I would like to use them in full scenario. Like user provide an URI and a method to read from URI. We do all the other things. What do you think? By the end, sea://, hdfs://, file:/// are URIs too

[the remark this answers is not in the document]
yangxk1: So I think we can donot storing the member variable `String prefix` in graphInfo/vertexInfo/edgeInfo, but using `URI baseURI`

maven-projects/info/src/main/java/org/apache/graphar/util/PathUtil.java
```diff
@@ -19,9 +19,16 @@
...
+
 public class PathUtil {
+
+    public static String resolvePath(String basePath, String subPath) {
+        return Path.of(basePath).resolve(subPath).toString();
+    }
+
     public static String pathToDirectory(String path) {
-        if (path.startsWith("s3://")) {
+        if (path.startsWith("s3://") || path.startsWith("oss://")) {
```
SemyonSinchenko, line 31: We can get the protocol name simply by `URI.getScheme()` as well, as we can get path by `URI.getPath()`

maven-projects/info/src/main/java/org/apache/graphar/info/loader/BaseGraphInfoLoader.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public abstract GraphInfo loadGraphInfo(URI graphYamlUri) throws IOException;
+
+    public abstract VertexInfo loadVertexInfo(URI vertexYamlUri) throws IOException;
+
+    public abstract EdgeInfo loadEdgeInfo(URI edgeYamlUri) throws IOException;
+
+    public GraphInfo buildGraphInfoFromGraphYaml(URI baseUri, GraphYaml graphYaml)
+            throws IOException {
+        String prefix = PathUtil.pathToDirectory(baseUri.getPath());
```
SemyonSinchenko, line 47: Why not pass URI to the PathUtil directly? It will allows us to get scheme (`file:///`, `ftp://`, `s3a://`, `hdfs://`, etc.) instead of checking it by `startsWith`. I think relying on the Java built-in API will make it more robust and flexible, compared to raw strings manipulations when we should cover all the corner cases...

maven-projects/info/src/test/java/org/apache/graphar/info/LocalFileSystemStringStreamLoader.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.apache.graphar.info;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import org.apache.graphar.info.loader.StreamGraphInfoLoader;
+
+public class LocalFileSystemStringStreamLoader extends StreamGraphInfoLoader {
```
SemyonSinchenko, line 28: What do you think about moving it to `org.apache.graphar.info.loader.impl`? I mean it can be useful, especially for unit tests.

maven-projects/info/src/test/java/org/apache/graphar/info/GraphLoaderTest.java
SemyonSinchenko, line 1: What do you think about adding tests for resolving for common schemes? Just to check that our resolving is working fine. Like `s3://`, `hdfs://`, `o3fs://`, `ofs://`, `gs://`, etc.? Not reading but just resolving. That for the most common protocols, we will be sure, that resolving of relative paths is working fine.

SemyonSinchenko (commented):
I like it more and more! Thanks a lot @yangxk1 ! I left a few minor comments.

yangxk1:
I added the test of `/tmp/`  `file:///tmp/` `s3://tmp/` `hdfs://tmp/`, and I think uri resolving will work fine as long as path meets the format (xxx://yyy)

SemyonSinchenko (approved):
I like it! Thanks @yangxk1 !

PR #754, opened 2025-09-10
feat(c++): Add support for the date/timestamp property types in iter

gary-cloud:
### Reason for this PR

Previously, `vertexIter` and `edgeIter` did not properly support `date32` and `timestamp(MILLI)` property types. 

* Added explicit handling for `arrow::date32` and `arrow::timestamp(MILLI)` types in `GetArrowArrayData`.
* Kept the existing default branch to raise a `Status::TypeError` if the array type is unsupported.

### Are these changes tested?

Yes

Yes, user-facing improvement:  Users can now reliably access `date32` and `timestamp(MILLI)` properties through iterators (`vertexIter` / `edgeIter`). No breaking changes introduced; unsupported types still return a `TypeError` as before.
```c++
// sample code for timestamp type
it.property<int64_t>("creationDate").value();
// sample code for date type
it.property<int32_t>("creationDate").value();
```

### DIFF
--- a/cpp/src/graphar/util.cc
+++ b/cpp/src/graphar/util.cc
@@ -89,6 +89,12 @@ Result<const void*> GetArrowArrayData(
   } else if (array->type()->Equals(arrow::boolean())) {
     return reinterpret_cast<const void*>(
         std::dynamic_pointer_cast<arrow::BooleanArray>(array).get());
+  } else if (array->type()->Equals(arrow::date32())) {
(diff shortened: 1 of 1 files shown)


gary-cloud:
> Hi @gary-cloud , this work makes sense! could you add some test case for different types?
> 
> yeah, of course, I'll add test for the two types.

yangxk1 (approved):
LGTM~ Thank you!

PR #764, opened 2025-09-15
feat(java,info): add vertex info builder

Jarvx200:
### Reason for this PR

Refactor for `VertexInfo` telescoping constructor ( #740 )

Added a simple builder for the class with basic testing

### Are these changes tested?

yes

no

### DIFF
--- a/maven-projects/info/src/test/java/org/apache/graphar/info/VertexInfoTest.java
+++ b/maven-projects/info/src/test/java/org/apache/graphar/info/VertexInfoTest.java
@@ -45,7 +45,8 @@ public class VertexInfoTest {
 
     @Test
     public void testVertexInfoBuilderDoubleDeclaration() throws URISyntaxException {
-        VertexInfo.VertexInfoBuilder doubleDefinitionBuilder = defaultBuilder.baseUri(new URI("world"));
+        VertexInfo.VertexInfoBuilder doubleDefinitionBuilder =
+                defaultBuilder.baseUri(new URI("world"));
         VertexInfo v = doubleDefinitionBuilder.build();
 
         Assert.assertEquals(new URI("world"), v.getBaseUri());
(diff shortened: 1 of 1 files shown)


yangxk1:
> ❌ Patch coverage is 68.96552% with 9 lines in your changes missing coverage. Please review.

Hi @Jarvx200 , pls add more test case

maven-projects/info/src/main/java/org/apache/graphar/info/VertexInfo.java
```diff
@@ -34,6 +34,78 @@ public class VertexInfo {
...
+            this.type = type;
+            return this;
+        }
+
+        public VertexInfoBuilder chunkSize(long chunkSize) {
+            this.chunkSize = chunkSize;
+            return this;
+        }
+
+        public VertexInfoBuilder propertyGroups(PropertyGroups propertyGroups) {
```
yangxk1, line 60: Add an `addPropertyGroup` method, just like in `EdgeInfoBuilder`

Thespica:
> Hey there, I've looked at the builders and there seem to be some recurring attributes that need to have specific methods and transient fields (eg. `PropertyGroups`), would it be ok if i create an abstract generic super-class for the builders so I keep the code concise and avoid redundancy?

Hi @Jarvx200 , that's an excellent suggestion. Creating an abstract super-class for the builders is a great approach to handle recurring attributes (like PropertyGroups, baseUri and etc.) and keep the code concise by avoiding redundancy.

Please feel free to proceed with that implementation. It's a solid design choice.

Jarvx200:
Sure thing, I will adhere this design choice and continue working, thanks for the feedback!

Jarvx200:
Looking on the errors it seems like the test fail because I extracted `PropertyGroups` from `PropertyGroup`. The problem with it being inside is the fact that the AbstractBuilder can't see it.

If it's ok, I think extracting the AbstractBuilder from it's package and placing it with the other classes which require it will help with this but also help with limiting the access modifiers in it's internal api.

Jarvx200:
Hey there, thanks a lot for the feedback. Looking back it seems a bit too complicated and it might be harder to mentain. I will try to find a middleground so I respect DRY principles but don't overcomplicate the codebase. 
Maybe embedding a small builder as an instance variable for common fields could be a bit of a better idea?

yangxk1:
@Jarvx200 ,We can implement it in a simple way, and then improve it later. What do you think?
We can temporarily add common files in both classes builder

Jarvx200:
For sure, I will do it like you proposed at first so it's working and then we can work on reducing redundancy!
@yangxk1

Issue #774, opened 2025-09-26
Reduce copies in graph_reader.h


Sober7135:
### Reduce copies in graph_reader.h

1. using `std::vector<IdType> filtered_ids` with `std::move` instead of `const std::vector<IdType> filtered_ids`
https://github.com/apache/incubator-graphar/blob/b7768ad09ece6941a32e65e436f765fdf3c1a0fc/cpp/src/graphar/high-level/graph_reader.h#L330-L333

2. using `const std::vector<std::string> &` instead of `std::vector<std::string> filter_labels`, since we never modify or own the `filter_labels`
https://github.com/apache/incubator-graphar/blob/b7768ad09ece6941a32e65e436f765fdf3c1a0fc/cpp/src/graphar/high-level/graph_reader.h#L379-L384

3. using `const std::string &` instead of `std::string`, since we never modify or own it.
https://github.com/apache/incubator-graphar/blob/b7768ad09ece6941a32e65e436f765fdf3c1a0fc/cpp/src/graphar/high-level/graph_reader.h#L386-L388

I’m willing to submit a PR for this.

### Component(s)

C++

PR #777, opened 2025-09-27
chore(c++): reduce copies in graph_reader.h and remove unnecessary `std::move`

Sober7135:
### Reason for this PR
related to #774 

1. using `const &` and `std::move` to reduce copies
2. remove unnecessary `std::move`

### Are these changes tested?

yes

no


yangxk1 (approved):
LGTM~ Thank you!

PR #819, opened 2025-12-29
[C++][CMake] Fix install/export targets and Windows runtime handling

### LINKED ISSUE #810: feat(c++): Improve CMakeLists.txt for portable library distribution and installation

rel: #806 
The current CMakeLists.txt in GraphAr does not fully support robust distribution and installation of the library. This issue proposes a comprehensive update to:

- Generate install targets for library components and headers
- Export CMake config files for `find_package` support in downstream projects
- Ensure headers, artifacts, and dependencies are correctly installed
- Facilitate use in various build systems and package managers

These enhancements will improve the adoption and packaging of GraphAr for both users and external contributors.

### Component(s)

C++

tanmoyPaul12:
Fixes #810

### Reason for this PR

The current CMake installation logic in GraphAr does not reliably support
library distribution and downstream consumption, especially on Windows.

This PR addresses the CMake install and export issues discussed in #810
and makes runtime installation behavior consistent across platforms.

- Remove duplicate and conflicting install logic for the `graphar` target
- Define a single canonical `install(TARGETS graphar ...)` rule
- Ensure headers and CMake package config files are installed correctly
- Fix Windows runtime (DLL) installation issues during `cmake --install`
- Validate installation with MSVC + vcpkg + Apache Arrow

### Are these changes tested?

- Windows 11
- MSVC (Visual Studio 2026)
- CMake + vcpkg
- Apache Arrow installed via vcpkg (arrow[dataset,parquet])
- Verified:
  - `cmake --build . --config Release`
  - `cmake --install . --config Release`
  - Generated headers and CMake config files

No. This change affects CMake install and packaging behavior only and does not
modify GraphAr APIs or runtime behavior.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -277,7 +277,7 @@ macro(build_graphar)
     else()
         add_library(graphar SHARED ${CORE_SRC_FILES})
     endif()
-    install_graphar_target(graphar)
+   # install_graphar_target(graphar)
     target_compile_features(graphar PRIVATE cxx_std_17)
     target_include_directories(graphar PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/thirdparty)
     target_link_libraries(graphar PRIVATE ${CMAKE_DL_LIBS})
@@ -334,7 +334,7 @@ macro(build_graphar_with_arrow_bundled)
     else()
         add_library(graphar SHARED ${CORE_SRC_FILES})
     endif()
-    install_graphar_target(graphar)
+    #install_graphar_target(graphar)
     target_compile_features(graphar PRIVATE cxx_std_17)
     target_include_directories(graphar PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/thirdparty)
     target_include_directories(graphar SYSTEM BEFORE PRIVATE ${GAR_ARROW_INCLUDE_DIR})
@@ -497,6 +497,14 @@ endif()
 # ------------------------------------------------------------------------------
 # Install
 # ------------------------------------------------------------------------------
+install(TARGETS graphar
+  EXPORT graphar-targets
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}   # .lib / .a
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}   # .so / .dylib
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} OPTIONAL  # .dll / executables
+)
+
+
 install(DIRECTORY ${PROJECT_SOURCE_DIR}/src/graphar
         DESTINATION include
         FILES_MATCHING


tanmoyPaul12:
@yangxk1 I’ll update this PR to:
- Ensure install(EXPORT ...) generates usable CMake target files
- Address the RPATH ($ORIGIN) and hybrid linking strategy discussed in the issue

Thanks for the clarification — I’ll follow up with the changes shortly.

keksmd (commented):
@tanmoyPaul12 thanks for chasing #810 — reliable install/export is a real gap, and the MSVC + vcpkg validation you did is exactly the kind of testing this area was missing.

Unfortunately I think this PR has been overtaken by `main` and needs a rebase before it can be reviewed on its merits. GitHub currently reports it as `CONFLICTING` / `DIRTY`.

## What changed underneath it

`refactor(build): restructure CMake files into subdirectories (#877)` (commit `d1fe7f9`) split `cpp/CMakeLists.txt` apart. The `build_graphar()` / `build_graphar_with_arrow_bundled()` macros and the `install_graphar_target()` helper that this PR edits no longer exist in `cpp/CMakeLists.txt`. The library and its install rule moved to `cpp/src/CMakeLists.txt`, which on current `main` reads:

```cmake
install(TARGETS graphar graphar_thirdparty
        EXPORT graphar-targets
        ARCHIVE DESTINATION lib
        LIBRARY DESTINATION lib
        RUNTIME DESTINATION bin
        INCLUDES
        DESTINATION include)
```

So the three headline items of this PR — removing the duplicate/conflicting install logic, defining a single canonical `install(TARGETS graphar ...)`, and adding a `RUNTIME` destination so the Windows DLL is installed — appear to already be in place. Before rebasing, it is probably worth re-testing your MSVC + vcpkg scenario against current `main` to see whether #810 is still reproducible at all.

## One thing that does still look wrong on `main`

The two halves of the install disagree about the library directory. `cpp/src/CMakeLists.txt` hardcodes `lib` and `bin`, while `cpp/CMakeLists.txt` places the package config files using the GNU variable:

```cmake
install(FILES "${PROJECT_BINARY_DIR}/graphar-config.cmake"
              "${PROJECT_BINARY_DIR}/graphar-config-version.cmake"
        DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/graphar)

install(EXPORT graphar-targets
        FILE graphar-targets.cmake
        DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/graphar)
```

`include(GNUInstallDirs)` is already present (`cpp/CMakeLists.txt`), and on distributions where `CMAKE_INSTALL_LIBDIR` is not plain `lib` (`lib64` on Fedora/RHEL, `lib/<triplet>` on Debian multiarch) the package config lands in one tree while the library lands in another. Making `cpp/src/CMakeLists.txt` use `${CMAKE_INSTALL_LIBDIR}`, `${CMAKE_INSTALL_BINDIR}` and `${CMAKE_INSTALL_INCLUDEDIR}` would be a small, self-contained follow-up, and a good scope for this PR after the rebase.

## Style note for whatever survives

The diff comments the old calls out rather than deleting them:

```cmake
-    install_graphar_target(graphar)
+   # install_graphar_target(graphar)
```

Commented-out code should be removed outright; the indentation also drops from four spaces to three on the first hunk, which `cmake-format` would flag.

PR #839, opened 2026-02-06
docs: add pre-commit guidelines and fetch test data via submodule

SYaoJun:


CONTRIBUTING.md
```diff
@@ -50,6 +50,18 @@ like (where issue \#42 is the ticket you're working on):
 ```shell
 $ git checkout -b 42-add-chinese-translations
 ```
+### Linting
```
yangxk1: Pre-commit does not only have the function of `Linting`, so it is more appropriate to call it `pre-commit`.

Issue #842, opened 2026-02-07
feat(C++): reduce copy for status.h


SYaoJun:
### Describe the enhancement requested

in incubator-graphar/cpp/src/graphar/status.h:157
> inline static Status OK() { return Status(); }
                                                        
this statement can return `{}` directly to avoid repeating the return type from the declaration

in incubator-graphar/cpp/src/graphar/status.h:138

> Status(StatusCode code, const std::string& msg) {
    state_ = new State;
    state_->code = code;
    state_->msg = msg;
  }

for msg can pass a value and use std::move 

### Component(s)

C++

PR #843, opened 2026-02-07
fix(c++): use std::move to reduce a copy

SYaoJun:
### Reason for this PR

I found a potential improvemnt spot in status.h. issue: #842


SYaoJun:
@yangxk1y could you please review my code

Issue #848, opened 2026-02-09
feat(C++): There are lots of compile warnings


SYaoJun:
### Describe the enhancement requested

I conceived two steps to fix those warning.

- [x]  Fix current compile warnings gradually #849 
- [x]  Add compile flags(`-Werror -Wextra` ) to CMakelists.txt #863 

### Component(s)

C++

SYaoJun:
<img width="1264" height="502" alt="Image" src="https://github.com/user-attachments/assets/34c793b1-4a6a-4bed-b271-5439bb299ef3" />

<img width="1187" height="511" alt="Image" src="https://github.com/user-attachments/assets/a8fc673f-0717-449f-9390-0c28ff04c028" />

<img width="1275" height="684" alt="Image" src="https://github.com/user-attachments/assets/e56e8bca-fa4f-4eb2-904f-d55e4dd230ad" />

PR #849, opened 2026-02-09
fix(c++): fix compile warnings(comparison of int and  uint,  unused result

SYaoJun:
…esult, catch value)

### Reason for this PR

Fix compile warnings.
issue: #848


SYaoJun:
@Sober7135   Would you mind helping me review the code? Yang's feedback has been incorporated. CC @yangxk1

yangxk1:
I see, Thank you~

PR #853, opened 2026-02-11
chore(C++): update the arrow version to latest 23.0.0

SYaoJun:
### Reason for this PR

The latest version of Apache Arrow is [23.0.0](https://github.com/apache/arrow/releases/tag/apache-arrow-23.0.0). This new version offers improved efficiency and numerous optimized features that would benefit us. How about updating the dependency to the latest version?


SYaoJun:
> Good job! Howere why do the cpp checks have been skipped

CI may be triggered when cpp files changed. I can pull again later.

PR #857, opened 2026-02-13
feat(rust): Enable Rust build on macOS and update Arrow linking

SYaoJun:
As discussions: https://github.com/apache/incubator-graphar/discussions/844

On macOS, the Rust bindings couldn't find the Arrow path. So I implemented a generic solution that works on both macOS and Ubuntu.

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -21,12 +21,13 @@ use std::env;
 use std::path::{Path, PathBuf};
 
 fn link_libraries() {
-    // TODO: Link to system Arrow, `libarrow` is under `/usr/lib/x86_64-linux-gnu/` on Ubuntu
-    println!("cargo:rustc-link-search=native=/usr/lib/x86_64-linux-gnu/");
-    println!("cargo:rustc-link-lib=dylib=arrow_compute");
-    println!("cargo:rustc-link-lib=dylib=arrow_dataset");
-    println!("cargo:rustc-link-lib=dylib=arrow_acero");
-    println!("cargo:rustc-link-lib=dylib=arrow");
+    println!("cargo:rerun-if-env-changed=PKG_CONFIG_PATH");
+
+    let _arrow = pkg_config::Config::new()
+        .statik(false)
+        .cargo_metadata(true)
+        .probe("arrow")
+        .expect("Arrow development files not found via pkg-config. Set PKG_CONFIG_PATH if needed.");
 
     println!("cargo:rustc-link-lib=graphar");
 }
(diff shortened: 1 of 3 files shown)


Sober7135:
Nice work! A couple of suggestions:

1. We could use the `pkg-config` crate instead of invoking `process::Command` directly: https://crates.io/crates/pkg-config
2. We should try `env::var("ARROW_LIB_DIR")` first. If it’s not set, then fall back to `pkg-config`. That seems more reasonable.

Sober7135:
Actually, I’m not very familiar with dependency handling, so this is a bit of an immature/personal suggestion.

Maybe we should rely fully on pkg-config, rather than `ARROW_HOME` and `ARROW_LIB_DIR`, since:
- Arrow always provides a *.pc file.
- We can get `CMAKE_PREFIX_PATH` via `pkg-config --variable=prefix arrow` (in Rust: `pkg_config::get_variable("arrow", "prefix")`).
- We can link Arrow via: `pkg_config::Config::new().statik(true).cargo_metadata(true)`

If the `*.pc` file isn’t in a standard location (e.g. `/usr/share/pkgconfig`), we can set `PKG_CONFIG_PATH`.

For example, if Arrow is installed at `/opt/arrow` and the `pkgconfig` files are under `/opt/arrow/lib64/pkgconfig`:
```bash
export PKG_CONFIG_PATH="/opt/arrow/lib64/pkgconfig:${PKG_CONFIG_PATH}" # set `PKG_CONFIG_PATH`
pkg-config --variable=prefix arrow
# output: /opt/arrow
```

Any thoughts?

rust/build.rs
```diff
@@ -63,6 +64,13 @@ fn build_graphar() -> Vec<PathBuf> {
         .define("CMAKE_BUILD_TYPE", cmake_build_type)
         .define("GRAPHAR_BUILD_STATIC", "ON")
         .define("GRAPHAR_ENABLE_SANITIZER", "OFF");
+
+    let arrow_prefix = pkg_config::get_variable("arrow", "prefix").ok();
+
+    if let Some(prefix) = arrow_prefix {
+        build.define("CMAKE_PREFIX_PATH", prefix);
+    }
```
Sober7135:
```Rust
if let Ok(prefix) = pkg_config::get_variable("arrow", "prefix") {
    build.define("CMAKE_PREFIX_PATH", prefix);
}
```
For simplicity, this is better.

Sober7135 (approved):
Thank you! LGTM!

SYaoJun:
> BTW, I’d suggest avoiding force-pushes, since rewriting history can make it harder for maintainers to track what changed (and it may invalidate existing review comments).

good suggestion. I will avoid forcing push code.

PR #858, opened 2026-02-13
fix(C++): remove redundant `inline` keyword

SYaoJun:
### Reason for this PR

- **member functions defined within a class declaration are implicitly inline** — explicit `inline` specifier here is redundant and unnecessary. (Ref: https://eel.is/c++draft/class.mfct#1)

- Adding `inline` to such functions does not change behavior, but adds unnecessary verbosity to the code. For consistency and readability, we remove redundant `inline` specifiers from in-class defined member functions in this PR.

### Are these changes tested?

yes

no


SYaoJun:
For the sake of simplicity and clarity, I’ve fixed the typos in another PR #875.

yangxk1 (approved):
LGTM

Issue #860, opened 2026-02-14
[Proposal][C++]: Add LRU chunk cache to Arrow chunk readers to avoid redundant file I/O


SYaoJun:
## Background

<img width="750" height="335" alt="Image" src="https://github.com/user-attachments/assets/51fdcba8-f3de-47ae-bf6f-0ab0ec7a8e6f" />

Currently, all Arrow chunk readers (
`VertexPropertyArrowChunkReader`, `AdjListArrowChunkReader`, `AdjListOffsetArrowChunkReader`, `AdjListPropertyArrowChunkReader`) discard the loaded `chunk_table_ `every time the chunk position changes via `seek(), next_chunk(), or seek_chunk_index().` This means that if a user seeks back to a previously loaded chunk, the entire Parquet file must be re-opened, metadata parsed, and data decoded again — even though the data hasn't changed.

This is particularly costly in graph traversal workloads (BFS, PageRank, label filtering) where vertex/edge access patterns exhibit strong locality, causing the same chunks to be read repeatedly.

## Proposal

Introduce a generic` LruCache<Key, Value> `and integrate it into all four chunk reader classes. When a chunk is loaded from disk, it is stored in the cache. On subsequent seeks to the same chunk, the cached `arrow::Table `is returned directly, avoiding file I/O entirely.

### Component(s)

C++

PR #861, opened 2026-02-14
feat(c++): add LRU chunk cache to Arrow chunk readers to avoid redundandt file IO

SYaoJun:
issue: #860


yangxk1 (commented):
Looking forward to a benchmark

PR #628, opened 2024-09-12
fix(c++): fix link error -lgar_arrow_static not found when link static graphar build arrow from source

acezen:
### Reason for this PR

as issue #627 describes, the gar_arrow_static symbol is not valid when the graphar building is done.
consider to link to the $GAR_ARROW_STATIC_LIB, instead gar_arrow_static

### What changes are included in this PR?

Change to link $GAR_ARROW_STATIC_LIB

### Are these changes tested?

yes

### Are there any user-facing changes?

no

PR #864, opened 2026-02-15
chore(scripts): add compile scripts in ubuntu and macos

SYaoJun:
### Reason for this PR

Every time I compile the C++ project, I have to input a long list of compilation options, which is time-consuming and completely unnecessary.

Regarding the legacy issue #628:

On macOS, we must explicitly specify the Arrow include path; otherwise, the compilation will fail.

Add two scripts on Ubuntu and MacOS

### Are these changes tested?

yes

no


yangxk1:
Hi @SYaoJun , thank you for doing this. Imo, it would be better to introduce the role of the script synchronously in readme/docs

PR #867, opened 2026-02-16
fix(c++): Fix unused header files and preallocate vector capacity

SYaoJun:


yangxk1:
I think we can accept copilot's suggestion before merging @SYaoJun

PR #877, opened 2026-02-25
refactor(build): restructure CMake files into subdirectories

SYaoJun:
### Reason for this PR

Currently, the CMake configuration is centralized in a single file, which makes it difficult to maintain and extend. A better approach is to modularize the configuration by splitting it into the corresponding subdirectories (e.g., examples, tests, benchmarks, and src) to improve maintainability, scalability, and separation of concerns.

move the cmake to subdirectories.


yangxk1 (approved):
Modifying the cmake structure brings great risks. I think we will continue to pay attention to CI and issue feedback in the next few weeks~

PR #878, opened 2026-02-25
refactor(cpp): replace int and -1 sentinels in graph info APIs

### LINKED ISSUE #865: [C++] Replace `int` in public metadata APIs with explicit types
## Background

The current public metadata APIs expose multiple count/index/version fields typed as `int`.

* The width of `int` is platform-dependent.
* Some lookup APIs rely on `-1` as a sentinel value.

## Candidate APIs to update (metadata layer)

  - `VertexInfo::PropertyGroupNum()`
  - `VertexInfo::GetPropertyGroupByIndex(int index)`
  - `EdgeInfo::PropertyGroupNum()`
  - `EdgeInfo::GetPropertyGroupByIndex(int index)`
  - `GraphInfo::GetVertexInfoIndex(...)`
  - `GraphInfo::GetEdgeInfoIndex(...)`
  - `GraphInfo::VertexInfoNum()`
  - `GraphInfo::EdgeInfoNum()`
  - `GraphInfo::GetVertexInfoByIndex(int index)`
  - `GraphInfo::GetEdgeInfoByIndex(int index)`
  - `InfoVersion` constructor / version-related APIs using raw `int`

## Proposed solution

Replace `int` with a fixed-width integer type or `size_t`.

However, this may be API-breaking, and there are multiple candidates:

### For count/index fields

* `size_t` — aligns with the C++ standard library (e.g., container `size()` and `operator[]`).
* `int32_t`
* `uint32_t`
* …

### For version fields (`InfoVersion`)

* `int32_t` or `uint32_t` should be sufficient.

## Compatibility / migration

Do we need to deprecate the existing functions first and introduce a new `v2` API?

### Component(s)

C++

shivendra-dev54:
### Reason for this PR
Fixes #865.

The standard int type is platform-dependent, and using -1 as a "not found" sentinel value leads to dangerous integer underflow bugs when comparing or casting against unsigned types like size_t (e.g., returning 18446744073709551615). This PR updates the GraphInfo APIs to use size-aware index types and safer missing-value handling.

- Replaced int with size_t for all container indices and sizes inside VertexInfo, EdgeInfo, and GraphInfo implementations.
- Updated LookupKeyIndex and associated index getter methods (like GetVertexInfoIndex) to return std::optional<size_t> instead of relying on -1.
- Removed unsafe static_cast<size_t>(-1) logic from search loops (e.g., inside RemovePropertyGroup and RemoveAdjacentList).

### Are these changes tested?
Yes, these changes are fully covered by the existing test suite:
 - Compiled successfully locally (make -j).
 - Passed local C++ test suite via ctest with GAR_TEST_DATA set (100% tests passed).
 - Code formatted using the devcontainer's clang-format.

Yes, for developers directly consuming the C++ APIs.

**BREAKING CHANGE: C++ API Return Types**

The following C++ APIs have been updated to use `size_t` or `std::optional<size_t>`:
- GraphInfo:
	- `GetVertexInfoIndex(const std::string& type)` -> Now returns `std::optional<size_t>`.
	- `GetEdgeInfoIndex(const std::string& src_type, const std::string& edge_type, const std::string& dst_type)` -> Now returns `std::optional<size_t>`.
	- `GetVertexInfoByIndex(size_t index)` -> Parameter changed from `int`.
	- `GetEdgeInfoByIndex(size_t index)` -> Parameter changed from `int`.
	- `VertexInfoNum()` -> Now returns `size_t`.
	- `EdgeInfoNum()` -> Now returns `size_t`.
- VertexInfo:
	- `PropertyGroupNum()` -> Now returns `size_t`.
	- `GetPropertyGroupByIndex(size_t index)` -> Parameter changed from `int`.
- EdgeInfo:
	- `PropertyGroupNum()` -> Now returns `size_t`.
	- `GetPropertyGroupByIndex(size_t index)` -> Parameter changed from `int`.

The following Rust APIs (graphar-rs) have been updated to use usize:
- VertexInfo:
	- `property_group_num(&self)` -> Now returns `usize`.
	- `get_property_group_by_index(&self, index: usize)` -> Parameter changed from `i32`.
- EdgeInfo:
	- `property_group_num(&self)` -> Now returns `usize`.
	- `get_property_group_by_index(&self, index: usize)` -> Parameter changed from `i32`.

### DIFF
--- a/cpp/src/graphar/graph_info.cc
+++ b/cpp/src/graphar/graph_info.cc
@@ -324,18 +324,18 @@ Result<std::string> VertexInfo::GetVerticesNumFilePath() const {
 }
 
 size_t VertexInfo::PropertyGroupNum() const {
-  return size_t(impl_->property_groups_.size());
+  return impl_->property_groups_.size();
 }
 
 std::shared_ptr<PropertyGroup> VertexInfo::GetPropertyGroup(
     const std::string& property_name) const {
   auto i = LookupKeyIndex(impl_->property_name_to_index_, property_name);
-  return !i.has_value() ? nullptr : impl_->property_groups_[i.value()];
+  return i.has_value() ? impl_->property_groups_[i.value()] : nullptr;
 }
 
 std::shared_ptr<PropertyGroup> VertexInfo::GetPropertyGroupByIndex(
     size_t index) const {
-  if (index >= size_t(impl_->property_groups_.size())) {
+  if (index >= impl_->property_groups_.size()) {
     return nullptr;
   }
   return impl_->property_groups_[index];
@@ -419,20 +419,15 @@ Result<std::shared_ptr<VertexInfo>> VertexInfo::RemovePropertyGroup(
   if (property_group == nullptr) {
     return Status::Invalid("property group is nullptr");
   }
-  std::optional<size_t> idx = std::nullopt;
   for (size_t i = 0; i < impl_->property_groups_.size(); i++) {
     if (*(impl_->property_groups_[i]) == *property_group) {
-      idx = i;
-      break;
+      return std::make_shared<VertexInfo>(
+          impl_->type_, impl_->chunk_size_,
+          RemoveVectorElement(impl_->property_groups_, i), impl_->labels_,
+          impl_->prefix_, impl_->version_);
     }
   }
-  if (idx == std::nullopt) {
-    return Status::Invalid("property group not found");
-  }
-  return std::make_shared<VertexInfo>(
-      impl_->type_, impl_->chunk_size_,
-      RemoveVectorElement(impl_->property_groups_, idx.value()), impl_->labels_,
-      impl_->prefix_, impl_->version_);
+  return Status::Invalid("property group not found");
 }
 
 bool VertexInfo::IsValidated() const { return impl_->is_validated(); }
@@ -743,7 +738,7 @@ std::shared_ptr<AdjacentList> EdgeInfo::GetAdjacentList(
 }
 
 size_t EdgeInfo::PropertyGroupNum() const {
-  return size_t(impl_->property_groups_.size());
+  return impl_->property_groups_.size();
 }
 
 const PropertyGroupVector& EdgeInfo::GetPropertyGroups() const {
@@ -758,7 +753,7 @@ std::shared_ptr<PropertyGroup> EdgeInfo::GetPropertyGroup(
 
 std::shared_ptr<PropertyGroup> EdgeInfo::GetPropertyGroupByIndex(
     size_t index) const {
-  if (index >= size_t(impl_->property_groups_.size())) {
+  if (index >= impl_->property_groups_.size()) {
     return nullptr;
   }
   return impl_->property_groups_[index];
@@ -923,21 +918,17 @@ Result<std::shared_ptr<EdgeInfo>> EdgeInfo::RemovePropertyGroup(
   if (property_group == nullptr) {
     return Status::Invalid("property group is nullptr");
   }
-  std::optional<size_t> idx = std::nullopt;
   for (size_t i = 0; i < impl_->property_groups_.size(); i++) {
     if (*(impl_->property_groups_[i]) == *property_group) {
-      idx = i;
+      return std::make_shared<EdgeInfo>(
+          impl_->src_type_, impl_->edge_type_, impl_->dst_type_,
+          impl_->chunk_size_, impl_->src_chunk_size_, impl_->dst_chunk_size_,
+          impl_->directed_, impl_->adjacent_lists_,
+          RemoveVectorElement(impl_->property_groups_, i), impl_->prefix_,
+          impl_->version_);
     }
   }
-  if (idx == std::nullopt) {
-    return Status::Invalid("property group not found");
-  }
-  return std::make_shared<EdgeInfo>(
-      impl_->src_type_, impl_->edge_type_, impl_->dst_type_, impl_->chunk_size_,
-      impl_->src_chunk_size_, impl_->dst_chunk_size_, impl_->directed_,
-      impl_->adjacent_lists_,
-      RemoveVectorElement(impl_->property_groups_, idx.value()), impl_->prefix_,
-      impl_->version_);
+  return Status::Invalid("property group not found");
 }
 
 bool EdgeInfo::IsValidated() const { return impl_->is_validated(); }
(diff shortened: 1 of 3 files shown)


cpp/src/graphar/graph_info.cc
```diff
@@ -886,20 +887,20 @@ Result<std::shared_ptr<EdgeInfo>> EdgeInfo::RemoveAdjacentList(
...
+  if (idx == std::nullopt) {
     return Status::Invalid("adj list not found");
   }
   return std::make_shared<EdgeInfo>(
       impl_->src_type_, impl_->edge_type_, impl_->dst_type_, impl_->chunk_size_,
       impl_->src_chunk_size_, impl_->dst_chunk_size_, impl_->directed_,
-      RemoveVectorElement(impl_->adjacent_lists_, static_cast<size_t>(idx)),
+      RemoveVectorElement(impl_->adjacent_lists_, idx.value()),
       impl_->property_groups_, impl_->prefix_, impl_->version_);
 }
```
Sober7135: I think we should use something like this:

```C++
for (size_t i = 0; i < impl_->adjacent_lists_.size(); i++) {
  if (impl_->adjacent_lists_[i]->GetType() == adj_list->GetType()) {
    return std::make_shared<EdgeInfo>(
        impl_->src_type_, impl_->edge_type_, impl_->dst_type_, impl_->chunk_size_,
        impl_->src_chunk_size_, impl_->dst_chunk_size_, impl_->directed_,
        RemoveVectorElement(impl_->adjacent_lists_, idx.value()),
        impl_->property_groups_, impl_->prefix_, impl_->version_);
  }
}
return Status::Invalid("adj list not found");
```

Maybe this is better?

BTW, I think this function may have a bug. It only compares the `adjacent_list` type for equality — is that sufficient? CC @yangxk1

Sober7135:
This PR will also break the deprecated Java FFI and the Python bindings. After it is merged, I will open an issue to track the follow-up work.

cpp/src/graphar/graph_info.cc
```diff
@@ -322,19 +323,19 @@ Result<std::string> VertexInfo::GetVerticesNumFilePath() const {
   return BuildPath({impl_->prefix_}) + "vertex_count";
 }
 
-int VertexInfo::PropertyGroupNum() const {
-  return static_cast<int>(impl_->property_groups_.size());
+size_t VertexInfo::PropertyGroupNum() const {
+  return size_t(impl_->property_groups_.size());
```
Sober7135: I think we can use `impl_->property_groups_.size()` directly...

cpp/src/graphar/graph_info.cc
```diff
@@ -322,19 +323,19 @@ Result<std::string> VertexInfo::GetVerticesNumFilePath() const {
...
 }
 
 std::shared_ptr<PropertyGroup> VertexInfo::GetPropertyGroup(
     const std::string& property_name) const {
-  int i = LookupKeyIndex(impl_->property_name_to_index_, property_name);
-  return i == -1 ? nullptr : impl_->property_groups_[i];
+  auto i = LookupKeyIndex(impl_->property_name_to_index_, property_name);
+  return !i.has_value() ? nullptr : impl_->property_groups_[i.value()];
```
Sober7135: use `i.has_value()`

cpp/src/graphar/graph_info.cc
```diff
@@ -1290,46 +1286,48 @@ const std::unordered_map<std::string, std::string>& GraphInfo::GetExtraInfo()
...
-int GraphInfo::VertexInfoNum() const {
-  return static_cast<int>(impl_->vertex_infos_.size());
+size_t GraphInfo::VertexInfoNum() const {
+  return static_cast<size_t>(impl_->vertex_infos_.size());
 }
 
-int GraphInfo::EdgeInfoNum() const {
-  return static_cast<int>(impl_->edge_infos_.size());
+size_t GraphInfo::EdgeInfoNum() const {
+  return static_cast<size_t>(impl_->edge_infos_.size());
```
Sober7135: Please remove the remaining redundant casts and use `.size()` directly.

cpp/src/graphar/graph_info.cc
```diff
@@ -1290,46 +1286,48 @@ const std::unordered_map<std::string, std::string>& GraphInfo::GetExtraInfo()
...
 }
 
 const std::shared_ptr<VertexInfo> GraphInfo::GetVertexInfoByIndex(
-    int index) const {
-  if (index < 0 || index >= static_cast<int>(impl_->vertex_infos_.size())) {
+    size_t index) const {
+  if (index >= static_cast<size_t>(impl_->vertex_infos_.size())) {
```
Sober7135: Please remove the remaining redundant casts and use `.size()` directly.

cpp/src/graphar/graph_info.cc
```diff
@@ -1290,46 +1286,48 @@ const std::unordered_map<std::string, std::string>& GraphInfo::GetExtraInfo()
...
     return nullptr;
   }
   return impl_->vertex_infos_[index];
 }
 
-const std::shared_ptr<EdgeInfo> GraphInfo::GetEdgeInfoByIndex(int index) const {
-  if (index < 0 || index >= static_cast<int>(impl_->edge_infos_.size())) {
+const std::shared_ptr<EdgeInfo> GraphInfo::GetEdgeInfoByIndex(
+    size_t index) const {
+  if (index >= static_cast<size_t>(impl_->edge_infos_.size())) {
```
Sober7135: Please remove the remaining redundant casts and use `.size()` directly.

cpp/src/graphar/graph_info.cc
```diff
@@ -1290,46 +1286,48 @@ const std::unordered_map<std::string, std::string>& GraphInfo::GetExtraInfo()
...
   return LookupKeyIndex(impl_->vtype_to_index_, type);
 }
 
 std::shared_ptr<EdgeInfo> GraphInfo::GetEdgeInfo(
     const std::string& src_type, const std::string& edge_type,
     const std::string& dst_type) const {
-  int i = GetEdgeInfoIndex(src_type, edge_type, dst_type);
-  return i == -1 ? nullptr : impl_->edge_infos_[i];
+  auto i = GetEdgeInfoIndex(src_type, edge_type, dst_type);
+  return !i.has_value() ? nullptr : impl_->edge_infos_[i.value()];
```
Sober7135: Please refactor this.

shivendra-dev54:
@Sober7135 
can you please recheck the codes
I have applied all changes in most of the places

Issue #884, opened 2026-02-27
[CI] MacOS does not compile in parallel


SYaoJun:
### Describe the bug, including details regarding any error messages, version, and platform.

https://github.com/apache/incubator-graphar/actions/runs/22453190956/job/65027211522?pr=880

```shell
/Users/runner/work/_temp/39b35018-6c2f-4a92-9893-c365533574f6.sh: line 4: nproc: command not found
```

`nproc ` only applied on Linux. MacOS should use anthoer arguments.

```shell
$(sysctl -n hw.ncpu)
```

A better approache is use [Ninja](https://github.com/ninja-build/ninja), which default runs in parallel.

### Component(s)

Continuous Integration

PR #885, opened 2026-02-27
fix(ci): replace make with ninja.

SYaoJun:
### Reason for this PR

fix: #884


yangxk1 (commented):
Using `ninja` looks good! but I think we should also keep a `make` case to ensure that make works fine

.github/workflows/ci.yml
SYaoJun: @yangxk1 I kept the make for this position

PR #888, opened 2026-02-28
fix(cpp): Handle npos in PathToDirectory for S3 URIs without query strings

### LINKED ISSUE #881: `PathToDirectory` can throw for valid S3 URIs without `?`
> `PathToDirectory`: for `s3://...` paths without a '?' query string, `find_last_of('?')` returns `npos` and `path.substr(t)` will throw `std::out_of_range`. Handle the `npos` case explicitly (e.g., treat it as no suffix) to avoid crashing on valid S3 URIs.
> ```suggestion
>     std::string prefix;
>     std::string suffix;
>     if (t == std::string::npos) {
>       prefix = path;
>       suffix.clear();
>     } else {
>       prefix = path.substr(0, t);
>       suffix = path.substr(t);
>     }
> ``` 

 _Originally posted by @Copilot in [#878](https://github.com/apache/incubator-graphar/pull/878/changes#r2858598451)_

shivendra-dev54:
### Reason for this PR
Fixes #881 

`PathToDirectory` was crashing with a `std::out_of_range` exception when parsing valid S3 URIs that do not contain a query string (`?`). This happens because `find_last_of('?')` returns `std::string::npos`, which was being passed directly into `path.substr()`.

Added an explicit check for `std::string::npos` in `PathToDirectory`. When an S3 URI without a query string is processed, it now correctly identifies the entire path as the prefix and sets the suffix to an empty string, safely avoiding the out-of-bounds crash.

### Are these changes tested?
Yes, the C++ test suite was compiled and run locally, and all tests pass successfully. 

No.

**Critical Fix: Fixes a bug that causes a crash (`std::out_of_range` exception) when parsing valid S3 URIs without query strings.**


shivendra-dev54:
@yangxk1 
I have incorporated all the changes mentioned by the copilot 
But the python mac build is still failing just like the #906

shivendra-dev54:
@yangxk1 
please do review this if it meets the issue expectation or not

PR #891, opened 2026-03-01
refactor(C++): change FilterOptions parameter to be moved instead of const reference

SYaoJun:
### Reason for this PR

use std::move instead of const refernce


yangxk1:
Nice! I will merge it

PR #912, opened 2026-03-30
test: add test from string to large string

SYaoJun:


Thespica:
fix the clang-format pleas @SYaoJun

PR #913, opened 2026-03-30
fix: avoid using C-style pointers to prevent potential memory leaks

SYaoJun:
### Reason for this PR

current `Status` missing copy assignment operator, which may cause memory leak.
modern cpp guidelines suggent using smart pointer instead of C-type pointer.

In this PR, it uses `std::unique_ptr<State> state_;` to replace orignal  `State* state_;`

BTW, This memory leak was detected by `.clang-tidy`.

#898

### DIFF
--- a/cpp/src/graphar/status.h
+++ b/cpp/src/graphar/status.h
@@ -233,33 +233,23 @@ class Status {
   }
 
   /** Return true iff the status indicates success. */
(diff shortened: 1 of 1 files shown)


Thespica:
[cpp faild to buil](https://github.com/apache/incubator-graphar/actions/runs/23752326239/job/69197546374?pr=913)

pleas fix that.

yangxk1 (approved):

PR #915, opened 2026-04-10
fix: avoid using implicit new operator to manage memory

SYaoJun:
### Reason for this PR

include common functions under graphar namespace and replace implicit memory with vector and smart pointer.


yangxk1 (approved):
LGTM~

PR #916, opened 2026-04-11
test: add test for label filter case

SYaoJun:
### Reason for this PR

Add test for filtering vertices by gender property using expression
filters. Defines expected count constants for validation.


yangxk1 (approved):
LGTM

Issue #167, opened 2023-05-24
[C++] [Improvement] Support to write std::vector in VerticesBuilder/EdgeBuilder of the C++ library


lixueclaire:
**Is your feature request related to a problem? Please describe.**
As the first step of issue #48, we propose to add new methods in the writers of the GraphAr C++ library to support the use of `std::vector` as an input parameter and write the data to generate GraphAr data files.

**Describe the solution you'd like**
To extend the high-level writer of the C++ library, i.e. [VerticesBuilder](https://alibaba.github.io/GraphAr/reference/api-reference-cpp.html#_CPPv4N12GraphArchive7builder15VerticesBuilderE) and [EdgesBuilder](https://alibaba.github.io/GraphAr/reference/api-reference-cpp.html#_CPPv4N12GraphArchive7builder12EdgesBuilderE), we recommend to implement the following function, for adding a column of a property:

```
class VerticesBuilder/EdgesBuilder {
  ...
  Status AddPropertyColumn(const std::string &property, const std::vector &values);
  ...
}
```
For implementation, the `values` can be added in `Vertices` of the builder one by one, or it can be saved separately and processed when `Dump()` is called.

**Describe alternatives you've considered**
As the next step, we suggest modifying the low-level writers of the C++ library to support writing `std::vector`.

**Additional context**
This issue is a part of issue #48 and is a good first issue for beginners to get familiar with our C++ library.

amrSherif12:
Hello @lixueclaire,
Im a new contributor here and I found this issue interesting and a good start for getting to know the project.
I will try to make a PR as soon as possible.

PR #928, opened 2026-05-23
feat(c++): Support to write vector in VerticesBuilder/EdgeBuilder

amrSherif12:
### Reason for this PR
This is a PR that solve issue #167 which adds support to write std::vector in VerticesBuilder/EdgeBuilder

I added a method in edges_builder.h and vertices_builder.h that go over the 
These changes were tested in the test_builder.cc file and all tests passed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -107,3 +107,20 @@ repos:
         files: ^maven-projects/spark/
         pass_filenames: false
         require_serial: true
+
+      - id: cargo-fmt
+        name: Rust cargo fmt
+        entry: bash -c 'cd rust && cargo fmt --all'
+        language: system
+        files: ^rust/.*\.rs$
+        pass_filenames: false
+        require_serial: true
+
+      - id: cargo-clippy
+        name: Rust cargo clippy
+        entry: bash -c 'cd rust && cargo clippy --all-targets --all-features -- -D warnings'
+        language: system
+        files: ^rust/.*\.(rs|toml|lock)$
+        pass_filenames: false
+        require_serial: true
+        stages: [manual]


SYaoJun (changes requested):
## PR Review: AddPropertyColumn for Builders

### Summary

This PR adds `AddPropertyColumn` to `VerticesBuilder` and `EdgesBuilder` to batch-set a property across all vertices/edges from a `std::vector<std::any>`. The idea is useful, but there is a **critical correctness bug** in the `EdgesBuilder` implementation, plus several smaller issues.

---

### 🔴 Critical: `edges_builder.h` — undefined iteration order over `edges_`

`edges_` is declared as:

```cpp
std::unordered_map<IdType, std::vector<Edge>> edges_;
```

`std::unordered_map` has **no defined iteration order** — it varies by platform, compiler, and even between runs with the same binary. The new code:

```cpp
int value = 0;
for (auto& [id, edges] : edges_) {
  for (Edge& edge : edges) {
    edge.AddProperty(property, values[value++]);
  }
}
```

...silently assigns `values[i]` to an **arbitrary** edge depending on the hash map's internal order. The property data is effectively shuffled randomly. This is a **data corruption bug**, not just a style issue.

**Fix options (pick one):**

1. **Change the API contract:** document that `values` must follow chunk-index-major order (sorted by `vertex_chunk_index`, then by insertion order within each chunk), and sort the outer map before iterating:
   ```cpp
   std::map<IdType, std::vector<Edge>> edges_;  // ordered
   // or sort at call time
   ```

2. **Change the data structure:** replace `unordered_map` with `std::map` or `std::vector<std::vector<Edge>>` (indexed by `vertex_chunk_index`). This gives deterministic order for free.

3. **Change the API shape:** accept a structure that explicitly maps each edge to its property value (e.g., a flattened `std::vector<std::any>` documented as chunk-major order, or per-chunk vectors).

Option 2 (`std::map` or `std::vector`) is the simplest correct fix.

---

### 🟡 Moderate Issues

#### 1. `edges_builder.h`: `int value = 0` — wrong type

```cpp
int value = 0;  // ❌ int may overflow for large graphs
```
Should be `size_t` or `IdType` to match the index into `values`.

#### 2. `vertices_builder.h`: consider using `size_t` for the loop index

```cpp
for (size_t i = 0; i < vertices_.size(); i++) {
  vertices_[i].AddProperty(property, values[i]);
}
```
This is correct because `vertices_` is a `std::vector` (deterministic order). A minor suggestion: use `vertices_.size()` directly instead of relying on the caller to pass the right-sized vector — but the existing size check already covers this.

#### 3. Missing `const` on the method

Both new methods modify the builder's state (adding properties to edges/vertices), so `const` is correctly **not** used. This is fine — just flagging for awareness.

---

### 🟢 Minor / Suggestions

#### 1. Test coverage: only `int` type is tested

```cpp
std::vector<std::any> values(builder->GetNum());
for (size_t i = 0; i < values.size(); i++) {
  values[i] = i;  // only int
}
```
The property being added (`"prop1"`) — what is its registered type in the graph schema? If the schema says `int64`, then `int` works. But it would be good to test with `std::string`, `double`, etc., to ensure `std::any` type conversion works correctly in `Dump()`.

#### 2. API documentation: clarify order semantics for `VerticesBuilder`

For `VerticesBuilder::AddPropertyColumn`, the doc says:

> `values` vector of values which gets mapped to each vertex with size equal to the vertices collection

It should explicitly state: *"`values[i]` is mapped to the i-th vertex in insertion order"*. This is currently true (because `vertices_` is a `std::vector`), but without documentation, future maintainers won't know the contract.

#### 3. Consider `[[nodiscard]]` on `AddPropertyColumn`

The return value is a `Status` — if the caller ignores it, the property addition silently fails. Adding `[[nodiscard]]` (already used on other methods in these classes) would help.

---

### Verdict

**Request Changes** — the `unordered_map` iteration order bug in `EdgesBuilder::AddPropertyColumn` is a data correctness issue that must be fixed before merge. The `int value = 0` type issue is also worth fixing now.

After the fix, please also add a test case that round-trips through `Dump()` + re-load and asserts the property values are correct — this would have caught the iteration order bug.

PR #929, opened 2026-05-24
feat: add PR template checklist to prevent trivial CI failures

SYaoJun:
### Reason for this PR

Noticed that the contributor reviewed their own PR

Add PR template checklist.

### Are these changes tested?
No need to test

no

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -45,3 +45,10 @@ Please uncomment the line below (and provide explanation) if the changes fix eit
 We use this to highlight fixes to issues that may affect users without their knowledge. For this reason, fixing bugs that cause errors don't count, since those are usually obvious.
 -->
 <!-- **Critical Fix: <description>** -->
+
+## Checklist
+
+- [ ] I have performed a self-review of my own code.
+- [ ] I have formatted my own code using `make cpplint` before submitting when changed files are in the `cpp` directory.
+- [ ] I have performed `pre-commit run` before commit the changed files.
+- [ ] I have added tests to prove my changes are effective.


yangxk1 (commented):
### Overview
Adds a 4-item checklist section to `.github/PULL_REQUEST_TEMPLATE.md` so contributors are nudged to self-review, lint, run pre-commit, and add tests before requesting review.

### What looks good
- The intent is healthy: an explicit checklist reduces back-and-forth on lint/format/CI failures, and it gives reviewers a quick way to point contributors back to the rules.
- Implementation is minimal and isolated — a single template file, no behavioral change to CI.

### Suggestions

1. **Missing trailing newline.** The diff ends with `\ No newline at end of file`. Most lint/pre-commit setups (and POSIX text-file convention) want a trailing newline; please add one. This is exactly the kind of "trivial CI failure" the PR is trying to prevent.

2. **Item 2 wording — `make cpplint` is a *linter*, not a formatter.** Suggest:
   > - [ ] I have run `make cpplint` (and any project-specific formatter, e.g. `clang-format`) on changed files under `cpp/`.

   Worth confirming with the maintainers whether the canonical command is `make cpplint`, `make format`, or both — the checklist should match what CI actually runs so contributors can reproduce the failure locally.

3. **Item 4 ("added tests") is too absolute.** Some PRs (docs-only, this very PR, CI tweaks) legitimately don't need new tests, and the existing template already asks contributors to *explain* why. Suggest:
   > - [ ] I have added tests that prove my changes are effective, **or** explained in the "Are these changes tested?" section above why tests are not needed.

4. **Item 1 ("self-review") doesn't really prevent CI failures** — it's a behavior nudge. That's fine, but the PR title overclaims a bit. Consider retitling to something like `feat: add self-review checklist to PR template` so the scope matches the change.

5. **Optional polish:** group items so the order matches the developer workflow — self-review → pre-commit (which often runs lint/format itself) → cpplint/format → tests. As written, items 2 and 3 overlap if `pre-commit` already runs `cpplint`/`clang-format`; mention that explicitly so contributors don't think they need to run both.

### Risk
Low. Template-only change, no code or CI behavior touched. LGTM once the trailing newline and item-2/item-4 wording are addressed.

yangxk1 (approved):

PR #937, opened 2026-07-18
chore(go): bootstrap Go SDK module and CI

ZekiLiu:
### Reason for this PR

This PR starts the Go SDK work tracked by #828.

The Go SDK needs a standalone module layout and CI/tooling foundation before
the actual GraphAr metadata APIs are added. Keeping this bootstrap change small
also makes the follow-up info implementation easier to review as a stacked PR.

Refs #828.

### What changes are included in this PR?

This PR adds the initial Go SDK scaffold:

- create the `go/graphar` Go module
- add the umbrella `graphar` package documentation
- add Go development targets for formatting, vetting, linting, testing,
  coverage, and license checks
- add a GitHub Actions workflow for Go CI when files under `go/**` change
- update license configuration for Go source files

This PR intentionally does not add the `types` or `info` package implementation
yet. Those APIs are planned for the stacked follow-up PR.

### Are these changes tested?

The intended local and CI checks are:

- `cd go/graphar && go mod tidy`
- `cd go/graphar && go test ./...`
- `cd go/graphar && make ci`
- `pre-commit run --files .github/workflows/go.yaml go/README.md go/graphar/.golangci.yml go/graphar/Makefile go/graphar/doc.go go/graphar/go.mod go/graphar/go.sum`

### Are there any user-facing changes?

Yes. This introduces the new Go SDK module directory and its README, but it does
not change any existing GraphAr API behavior.

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have formatted my own code using `make cpplint` before submitting when changed files are in the `cpp` directory. (Not applicable: this PR does not change files in the `cpp` directory.)
- [x] I have performed `pre-commit run` before commit the changed files.
- [x] I have added tests to prove my changes are effective. (CI/tooling bootstrap is covered by the Go workflow and module checks listed above; implementation tests are added in the stacked follow-up PR.)

Issue #828, opened 2026-01-16
[Proposal] Implementation Plan for Native Go SDK


ZekiLiu:
### Describe the enhancement requested

Hello GraphAr community,

I would like to propose the creation of a native Go SDK for GraphAr and am willing to kickstart the implementation and maintain it.

Motivation

Go (Golang) is a dominant language in the cloud-native landscape, widely used for building high-performance backend services, data pipelines, and infrastructure tooling. Currently, for a Go application to interact with GraphAr data, it would need to rely on complex solutions like CGO bindings to the C++ library or inter-process communication with a Java/Spark service.

A native Go SDK would significantly lower the barrier to adoption for the vast Go ecosystem by providing an idiomatic, efficient, and dependency-free way to read and write GraphAr formatted data. This would enable direct integration with Go-based graph databases, analysis tools, and data processing frameworks.

Preliminary Design Proposal (Open for Feedback)

To ensure consistency and maintainability, the Go SDK's design will be heavily inspired by the architecture of the existing C++ and Java/Spark/Python/Rust libraries. The core idea is to follow the same layered approach:

info Package: Pure Go data structures that represent the GraphAr schema (GraphInfo, VertexInfo, EdgeInfo, PropertyGroup, etc.) and logic for parsing/serializing the .info.yml files. This will leverage the existing Proto definitions introduced in PR #573.

storage Package: An abstraction layer for accessing the underlying storage (e.g., local filesystem, S3), making the SDK storage-agnostic.

parquet Package: A dedicated module for handling Parquet file I/O, as it's the most common payload format. This will leverage a robust Go Parquet library (e.g., parquet-go or arrow/go).

reader Package: Provides high-level APIs to read vertex/edge chunks, handling different AdjList types and navigating through property groups.

writer Package: Provides high-level APIs to write vertex/edge data into the correct chunked and partitioned directory structure, including generating offset and metadata files.

Phased Development Plan

I propose tackling this in a phased approach to deliver value incrementally and gather feedback:

Phase 1: Core Schema & Reader (Read-Only)

- [ ] Implement the info package for YAML parsing and validation (based on generated Protos).

- [ ] Implement the storage layer with an initial local filesystem backend.

- [ ] Implement the parquet reader logic.

- [ ] Implement the reader package to read existing GraphAr vertex and edge data (all AdjList types).

- [ ] Add comprehensive unit tests using sample data generated by the Spark library to ensure compatibility.

Phase 2: Writer Implementation (Read-Write)

[...]

PR #941, opened 2026-07-27
feat(go,types): add FileType, Cardinality, AdjListType

ZekiLiu:
### Reason for this PR

Part of the pure-Go GraphAr SDK (tracking issus #828), following the bootstrap PR #937.
This is the foundational `types` package: the primitive value enums every higher
layer (info, reader/writer) depends on. Kept separate from `DataType` (next PR) so
each PR stays a single, reviewable module. The diff is a bit over the usual size,
but about half is table-driven tests and the four enums are one cohesive,
dependency-free unit; splitting them further would fragment the value layer.

New package `go/graphar/types` with four on-disk value types plus their string
parse/serialize and sentinel errors:

- `FileType` — csv / parquet / orc / json
- `Cardinality` — single / list / set
- `AdjListType` — unordered/ordered × by_source/by_dest, with helpers to convert
  to/from the legacy `(ordered, aligned_by)` form (`src`/`dst`)

Enum members and on-disk spellings match the C++ and Java SDKs (verified
member-by-member); Rust and Python are FFI/bindings over C++. No Arrow or
third-party dependency.

### Are these changes tested?

Yes. Table-driven unit tests cover round-trips, error paths, defensive copying,
and the lenient/strict version-string cases.

Yes — this adds the public `types` package. No breaking changes (new code only).

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -118,7 +118,9 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "stable"
+          # golangci-lint v1.64.8 cannot decode export data from newer Go
+          # stable releases; keep the lint job on the go.mod floor.
+          go-version: "1.23"
           cache-dependency-path: go/graphar/go.sum
 
       - name: golangci-lint
--- a/go/graphar/types/adjlist.go
+++ b/go/graphar/types/adjlist.go
@@ -1,0 +1,127 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package types
+
+import "fmt"
+
+// AdjListType is the layout of an edge type's adjacency list, combining
+// alignment side (source or destination vertex) with ordering (ordered or not).
+// Compare with ==; the numeric values carry no meaning.
+type AdjListType uint8
+
+// AdjListType values. The zero value is not a valid layout.
+const (
+	_ AdjListType = iota
+	// UnorderedBySource groups edges by source vertex, source side unordered
+	// (COO layout).
+	UnorderedBySource
+	// UnorderedByDest groups edges by destination vertex, dest side unordered.
+	UnorderedByDest
+	// OrderedBySource groups edges by source vertex, sorted by source (CSR).
+	OrderedBySource
+	// OrderedByDest groups edges by destination vertex, sorted by dest (CSC).
+	OrderedByDest
+)
+
+// alignedBySrc / alignedByDst are the on-disk aligned_by values, shared by
+// AlignedBy (write) and AdjListTypeFromOrderedAligned (read).
+const (
+	alignedBySrc = "src"
+	alignedByDst = "dst"
+)
+
+// On-disk spellings, shared by String (write) and ParseAdjListType (read) so the
+// two cannot drift.
+const (
+	adjListUnorderedBySourceName = "unordered_by_source"
+	adjListUnorderedByDestName   = "unordered_by_dest"
+	adjListOrderedBySourceName   = "ordered_by_source"
+	adjListOrderedByDestName     = "ordered_by_dest"
+)
+
+// String returns the on-disk spelling.
+func (a AdjListType) String() string {
+	switch a {
+	case UnorderedBySource:
+		return adjListUnorderedBySourceName
+	case UnorderedByDest:
+		return adjListUnorderedByDestName
+	case OrderedBySource:
+		return adjListOrderedBySourceName
+	case OrderedByDest:
+		return adjListOrderedByDestName
+	default:
+		return "unknown"
+	}
+}
+
+// IsOrdered reports whether the layout is ordered.
+func (a AdjListType) IsOrdered() bool {
+	return a == OrderedBySource || a == OrderedByDest
+}
+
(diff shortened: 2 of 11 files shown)


keksmd (commented):
@ZekiLiu this is nicely built — the shared unexported name constants so `String` and `Parse*` cannot drift, the defensive `slices.Clone` on both the constructor and the getters, and the sentinel errors wrapped with `%w` are all the right instincts for a package this foundational. CI is green across `Ubuntu Go 1.23`, `Ubuntu Go stable`, `golangci-lint` and `Pre-commit`.

I compared the semantics against the C++ and Java implementations, since `types` is where a divergence would silently propagate into every reader built on top of it. Three findings, one of which I think is worth fixing before merge.

## 1. An unknown schema version is accepted and then fails silently

`version2types` only knows version `1`, but nothing rejects other versions:

```go
func (v InfoVersion) CheckType(typeStr string) bool {
	if slices.Contains(version2types[v.version], typeStr) {
```

For `v.version == 2` the map lookup yields the nil slice, so `CheckType` returns `false` for `bool`, `int32`, `int64`, `float`, `double` and `string` alike. `ParseInfoVersion("gar/v2")` succeeds, `Validate()` passes, and the failure only surfaces much later as "every property type in this graph is unsupported".

Both other implementations refuse it at construction time. C++ (`cpp/src/graphar/version_parser.h`):

```cpp
explicit InfoVersion(int version) : version_(version) {
    if (version2types.find(version) == version2types.end()) {
      throw std::invalid_argument("Unsupported version: " + std::to_string(version));
    }
}
```

so `InfoVersion::Parse("gar/v2")` returns `Status::Invalid`. `maven-projects/info/.../VersionInfo.java` carries the same table and the same expectation.

Suggest having `ParseInfoVersion` (and ideally `Validate`) reject a version that is not a key of `version2types`, so a future `gar/v2` file fails loudly at the metadata boundary instead of degrading into a type-check failure.

## 2. The claimed C++ parity of `ParseInfoVersion` does not quite hold

The doc comment says:

> The tolerances match the C++ parser so both read the same on-disk strings.

I found two inputs where they disagree. The C++ side uses `std::regex_match` (anchored, whole-string) with:

```cpp
const std::regex version_regex("gar/v(\\d+).*");
const std::regex user_define_types_regex("gar/v\\d+ *\\((.*)\\).*");
```

**Leading whitespace.** `ParseInfoVersion` starts with `strings.TrimSpace(s)`, so `" gar/v1"` parses. `regex_match(" gar/v1", version_regex)` does not match, so C++ returns `Status::Invalid`. Java's `VersionParser` uses the identical pattern with `Matcher.matches()`, same result.

**Position of the type list.** The C++ pattern requires `(` to follow the digits with only spaces in between. Go takes the first `(` anywhere in the remainder:

```go
if lparen := strings.IndexByte(rest, '('); lparen >= 0 {
```

So `"gar/v1 xyz(a,b)"` yields `userDefinedTypes = ["a","b"]` in Go, and an empty type list in C++ and Java.

Neither is likely to appear in a real fixture, but since the comment promises parity it should either be made true or reworded to state where Go is deliberately more lenient. (For what it's worth `"gar/v0"` does agree — Go rejects it as non-positive, C++ rejects it as an unsupported version.)

## 3. Doc reference to code that does not exist yet

`filetype.go`:

```go
// FileTypeJSON is the JSON text format. Parseable for read-only fixture
// compatibility but rejected by PropertyGroup.Validate.
```

`PropertyGroup` is not part of this PR, so godoc renders a dangling reference. Either soften the wording now or land it together with the type it names.

## Smaller note

`NewInfoVersion` silently clamps a non-positive version to `DefaultVersion`, while `ParseInfoVersion` returns an error for the same conceptual input. That asymmetry also makes the `v.version <= 0` branch in `Validate` reachable only for the zero value. Returning an error from the constructor, or at least documenting the clamp as deliberate leniency for programmatic callers, would make the two paths easier to reason about.

None of this touches the `AdjListType` or `Cardinality` files, which read cleanly to me and match the on-disk spellings used by the other SDKs.

ZekiLiu:
@keksmd , thanks for the detailed review. I pushed a commit that addresses the feedback:

  1. Unsupported versions (e.g. gar/v2) are now rejected by ParseInfoVersion, Validate, and NewInfoVersion, so they fail at the metadata boundary instead of surfacing later as "unsupported property types".

  2. Parsing is now aligned with the C++/Java behavior: leading whitespace is rejected, and the type list is only parsed when '(' directly follows the version number (with spaces allowed in between). The doc
  comment was updated to match the actual behavior.

  3. The FileTypeJSON doc comment no longer references PropertyGroup.

  4. NewInfoVersion now returns an error instead of silently clamping non-positive versions, as you suggested.

  Would you mind taking another look when you have time?

keksmd (approved):
@ZekiLiu I re-read the package at `2c220c7` and checked out the branch to run the cases I originally raised, rather than reading the diff alone. All three findings plus the smaller note are resolved, and the parity now holds where it claimed to. LGTM.

What I verified locally (`go test ./...` in `go/graphar`, 24 tests, all passing on Go 1.26):

```
" gar/v1"           -> ERR must start with "gar/v"
"gar/v2"            -> ERR unsupported version 2
"gar/v0"            -> ERR unsupported version 0
"gar/v1 xyz(a,b)"   -> v=1 udt=[]        (list not taken from a detached paren)
"gar/v1\t(a,b)"     -> v=1 udt=[]        (tab is not a space, as in the C++ ` *`)
"gar/v1 (a,b)"      -> v=1 udt=[a b]
"gar/v1(a,b)"       -> v=1 udt=[a b]
"gar/v1 trailing"   -> v=1 udt=[]        (trailing text still tolerated, as in C++ `.*`)
"gar/v01"           -> v=1               (matches C++ stoi on the captured digits)
NewInfoVersion(2)   -> ERR unsupported version 2
NewInfoVersion(0)   -> ERR version must be positive
```

That is the same set of answers `std::regex_match` gives for the anchored `gar/v(\d+).*` and `gar/v\d+ *\((.*)\).*` patterns, so the doc comment's parity claim is now true rather than aspirational. Rejecting an unknown version in all three of `ParseInfoVersion`, `Validate` and `NewInfoVersion` also closes the path where a `gar/v2` file would have degraded into "every property type is unsupported" much further downstream. The `CheckType` rewrite to a comma-ok lookup is the right shape for the same reason.

One leftover, non-blocking: `DefaultVersion` is now referenced only by its own doc comment, since the clamp that used it is gone. It is still a reasonable thing to export for callers that need the assumed version of an Info object without a version field, so I would keep it — just flagging that a linter may eventually notice.

Nothing else from me. CI is green across all six checks.

PR #958, opened 2026-08-18
feat(java): add dependency-light storage API

keksmd:
### Reason for this PR

Implements focused issue #957, a dependency-light storage boundary for the pure-Java GraphAr roadmap. It is the API half of the broader storage work in #953 and deliberately excludes the local adapter, so this PR is independently reviewable and mergeable.

### What changes are included in this PR?

- Adds the `graphar-storage-api` module.
- Defines `Storage`, `InputFile`, `OutputFile`, `SeekableInput`, and `PositionOutput`.
- Specifies Java 11 `ByteBuffer` read/readFully, seek, EOF, and byte-position behavior.
- Adds direct and heap buffer tests for the seek/read contract.

### Are these changes tested?

Yes, on current `upstream/main`:

- `mvn --no-transfer-progress -P!nexus-download -pl storage-api -am spotless:check`
- `mvn --no-transfer-progress -P!nexus-download -pl storage-api -am clean verify -Dspotless.check.skip=true` — 3 tests, 0 failures/errors/skips.
- `pre-commit run --files` for all eight changed files and `git diff --check`.

### Are there any user-facing changes?

Yes. This adds a new optional Java module and public storage interfaces only. It does not alter existing GraphAr metadata or file-layout behavior.

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have formatted my own code using `make cpplint` before submitting when changed files are in the `cpp` directory.
- [x] I have performed `pre-commit run` before commit the changed files.
- [x] I have added tests to prove my changes are effective.

Issue #953, opened 2026-08-11
[FEATURE][Java] Add dependency-light storage API and local adapter


keksmd:
## Describe the enhancement requested

Add a dependency-light pure-Java storage boundary as the verified Phase 2 prerequisite for the Java reader/writer roadmap. The API isolates URI-backed byte access from GraphAr metadata/layout and from physical formats, so a future Parquet backend can consume files without importing Parquet, Hadoop, Arrow, or object-store dependencies into `graphar-info`.

Component(s): Java

## Scope

- `graphar-storage-api`: `Storage`, `InputFile`, `OutputFile`, `SeekableInput`, and `PositionOutput` contracts;
- `graphar-storage-local`: `file:` URI implementation with seek, bounded reads, create-vs-overwrite semantics, and byte positions;
- tests for heap/direct `ByteBuffer`, seek/readFully, overwrite safety, and unsupported URI rejection.

## Non-goals

No GraphAr layout resolution, metadata change, Parquet/ORC decoder, Hadoop/S3 adapter, Arrow public type, predicate/query API, reader, or writer.

## Compatibility and acceptance

The API must remain Java 11-compatible and add no runtime dependency to `graphar-info`. A local adapter must preserve byte content and positions, reject non-file URIs explicitly, and pass repository pre-commit plus the storage reactor's devcontainer `clean verify`.

## Dependencies and evidence

Relates to reader/writer umbrella #947 and architecture umbrella #756. Product-fork evidence is keksmd/incubator-graphar#3: `pre-commit run --files` passed and the declared devcontainer produced storage API/local artifacts with 7/7 JUnit tests.

Issue #947, opened 2026-08-11
[FEATURE][Java] Pure-Java GraphAr reader/writer implementation


keksmd:
## Describe the enhancement requested

Track the production-driven pure-Java data-access implementation as one umbrella. The product fork develops it immediately and independently; this Apache issue organizes only the small upstreamable slices.

Component(s): Java

## Proposed layering

`graphar-info` -> storage API -> IO request/capability API -> Parquet backend -> GraphAr chunk/layout reader -> ordered-by-source `neighbors` and `scanEdges` -> writer/validator.

## Non-goals for the first vertical

No JNI/FastFFI, mandatory Arrow representation, general query AST, Gremlin, graph algorithms, ORC/CSV, unordered layouts, cloud storage, or integrations.

## Acceptance for the first vertical

A Java reader resolves `vertexId -> offsets -> minimum Parquet range -> destinations` for `ordered_by_source`, exposes streaming edge scans, and preserves the dependency boundary. Writer, other layouts, and adapters are separate small PRs after the reader is proven.

## Links

Architecture umbrella: #756. Compatibility/TCK umbrella: #944. Open a storage, IO, Parquet, reader, or writer subissue only when its independently verified fork slice is ready to be proposed upstream.

PR #960, opened 2026-08-19
feat(java): let the storage API reach the local filesystem

keksmd:
### Reason for this PR

Closes the remaining half of #953. The storage API merged in #958 defines
`Storage`, `InputFile`, `OutputFile`, `SeekableInput`, and `PositionOutput`,
but no implementation ships with it, so nothing in the Java tree can open a
GraphAr file yet. Every later layer of #947 (physical IO, Parquet backend,
reader, writer) needs at least one working adapter to be testable.

A local filesystem adapter in a new `graphar-storage-local` module:

- `LocalStorage` resolves a `file:` URI (and a bare path) to a `java.nio.Path`
  and hands back input and output files.
- `LocalSeekableInput` reads at an absolute position and reports the end of the
  stream instead of a short buffer.
- `LocalOutputFile` / `LocalPositionOutput` create parent directories on demand,
  refuse an existing target unless the caller asked to replace it, and report
  the number of bytes written so far.

The module depends only on `graphar-storage-api`. It carries no GraphAr layout,
format, projection, or query concern, per the boundary agreed in #953.

### Are these changes tested?

Yes. `LocalStorageTest` covers the round trip, positional reads, end-of-stream
behaviour, directory creation, and the refusal to overwrite an existing target.

```
mvn --no-transfer-progress -pl storage-local -am clean verify
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

A new module and package `org.apache.graphar.storage.local`. Nothing existing
changes.

## Checklist

### DIFF
--- a/maven-projects/storage-local/src/main/java/org/apache/graphar/storage/local/LocalPositionOutput.java
+++ b/maven-projects/storage-local/src/main/java/org/apache/graphar/storage/local/LocalPositionOutput.java
@@ -29,6 +29,7 @@ final class LocalPositionOutput implements PositionOutput {
 
     private final OutputStream output;
     private long position;
+    private byte[] transfer;
 
     LocalPositionOutput(OutputStream output) {
         this.output = output;
@@ -48,11 +49,13 @@ final class LocalPositionOutput implements PositionOutput {
             return;
         }
 
-        byte[] buffer = new byte[Math.min(source.remaining(), BUFFER_SIZE)];
+        if (transfer == null) {
+            transfer = new byte[BUFFER_SIZE];
+        }
         while (source.hasRemaining()) {
-            int length = Math.min(source.remaining(), buffer.length);
-            source.get(buffer, 0, length);
-            write(buffer, 0, length);
+            int length = Math.min(source.remaining(), transfer.length);
+            source.get(transfer, 0, length);
+            write(transfer, 0, length);
         }
     }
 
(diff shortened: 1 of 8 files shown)


maven-projects/storage-local/src/main/java/org/apache/graphar/storage/local/LocalPositionOutput.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public void write(ByteBuffer source) throws IOException {
+        if (source.hasArray()) {
+            int length = source.remaining();
+            write(source.array(), source.arrayOffset() + source.position(), length);
+            source.position(source.position() + length);
+            return;
+        }
+
+        byte[] buffer = new byte[Math.min(source.remaining(), BUFFER_SIZE)];
```
SemyonSinchenko: Should we create and reuse the instance' own buffer (of the size `BUFFER_SIZE`) instead of the new allocation per call? It is synchronous (not thread safe dues to `position`), so I see no problem at all.

keksmd: Done in 3dde49e. The staging array is now one instance field, allocated on first use so an output that only ever sees heap-backed buffers never pays for it.

While writing it I found the direct-buffer path had no test that crossed a chunk boundary (the existing one wrote 2 bytes), so I added one for a 20000-byte direct buffer. With the copy loop reduced to a single pass it fails with `expected:<20000> but was:<8192>`, so it does hold the loop down.

SemyonSinchenko (approved):

SemyonSinchenko:
@SYaoJun @yangxk1 Hi! Let's merge this one? Java module is a work-in-progress so I see no problems at all.

Issue #959, opened 2026-08-19
[Feat][Java] Format-neutral physical batch and write contract


keksmd:
Part of #947.

The layering agreed in #947 puts a physical IO boundary between the GraphAr
reader/writer and any concrete file format. Before either vertical can be
built, the project needs one description of a row group that names no format:
a column type, a schema, a row, a batch, and a cursor over batches.

Scope:

- `ColumnType`, `Field`, `Schema` — neutral, recursive type description.
- `Row`, `RecordBatch`, `BatchCursor` — row access with one documented Java
  value mapping per column kind.
- `WriteRequest`, `WriteMode`, `PhysicalWriter` — the write half of the
  boundary, which shares every one of the types above.

Non-goals for this issue:

- The read request (projection, filter, row range, capability reporting) and
  `PhysicalReader`, which follow as their own issue against this base.
- Arrow as the public batch representation.
- Any GraphAr layout, chunking, or query semantics.

The module must not depend on any other GraphAr module.

PR #961, opened 2026-08-19
feat(java): describe physical batches and writes without a format

keksmd:
### Reason for this PR

Part of #947, tracked by #959. The reader and writer verticals both need one format-neutral description of a physical batch before either can name a file format. Without it, the Parquet backend and the GraphAr reader would each invent their own representation and the two would have to be reconciled later.

A new dependency-light `graphar-io-api` module holding a columnar physical-batch description plus the write half of the physical IO boundary:

- `ColumnType`, `Field`, `Schema` — recursive Arrow-shaped schema descriptions. Nested `LIST`, `FIXED_SIZE_LIST`, `STRUCT`, and `MAP` retain child field names/nullability; fixed binary and decimal parameters are explicit. Duplicate field names are valid because physical access is by position.
- `ValueVector`, `VectorRecordBatch`, `BatchCursor` — a batch is a schema, same-length vectors, and a row count. `VectorRecordBatch` rejects mismatched schema fields, vector width, and value counts before publication. It deliberately resembles Arrow's vector access shape without imposing Arrow buffers, allocators, or a transitive runtime dependency.
- `WriteRequest`, `WriteMode`, `PhysicalWriter` — a URI, a schema, a target disposition, and the writer that drains a cursor. `CREATE_NEW`, `OVERWRITE`, and `APPEND` are explicit.

A format adapter owns vector lifetime and the Java representation returned by `ValueVector#getObject`. That lets a Parquet or Arrow adapter expose native/lifecycle-bound data when safe, while keeping the base API usable without Arrow.

The read half of the boundary (`ReadRequest`, projection, filter, row range, capability reporting, `PhysicalReader`) is deliberately left out and will follow as its own PR against this base.

### Are these changes tested?

Yes — 33 tests. They cover value-object equality, indexed duplicate schema fields, recursive nested types, vector/batch structural validation, nullable list elements, vector-based cursor traversal, cursor ownership, schema mismatch, and all three write dispositions against an in-memory reference writer.

```
mvn --no-transfer-progress -f maven-projects/pom.xml -pl io-api -am verify
Tests run: 33, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

A new module and package `org.apache.graphar.io`. Nothing existing changes.

## Checklist


maven-projects/io-api/src/main/java/org/apache/graphar/io/Field.java
```diff
@@ -0,0 +1,53 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.graphar.io;
+
+import java.util.Objects;
+
+/** A named, typed physical column in a {@link Schema}. */
+public final class Field {
```
SemyonSinchenko, line 25: Checking a field in schema is a common op, let's add `equals` and `hash` as well.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Schema.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public Schema(List<Field> fields) {
+        Objects.requireNonNull(fields, "Schema fields cannot be null.");
+        List<Field> copy = new ArrayList<>(fields.size());
+        Set<String> names = new HashSet<>();
+        for (Field field : fields) {
+            Field nonNullField = Objects.requireNonNull(field, "A schema field cannot be null.");
+            if (!names.add(nonNullField.name())) {
+                throw new IllegalArgumentException(
+                        "Schema contains duplicate field: " + nonNullField.name());
```
SemyonSinchenko: Are we sure that it is a responsibility of the `Schema` to check for duplicates? Because fields can be accessed by index (and it is the only way to access), I see no problems in duplicated names tbh

maven-projects/io-api/src/main/java/org/apache/graphar/io/RecordBatch.java
```diff
@@ -0,0 +1,32 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.graphar.io;
+
+/** A format-neutral, finite group of rows sharing one schema. */
+public interface RecordBatch {
```
SemyonSinchenko, line 23: Why row-oriented model, not columnar? I mean we can have rows and record batch of them, or we can have a column (or vector in Apache Arrow terms) and a batch of them.

Tbh I would prefer to mimic the Apache Arrow layout (as follow as possible to what is described in the Apache Arrow Java). The GraphAr C++ already rely on Arrow.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ColumnType.java
```diff
@@ -0,0 +1,89 @@
...
+ * under the License.
+ */
+
+package org.apache.graphar.io;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/** A recursive, format-neutral type used in a physical {@link Schema}. */
+public final class ColumnType {
```
SemyonSinchenko, line 28: Do we want a more rich hierarchy? FixedSized, VariableSized, Complex (List, etc.), etc.? This one looks complete but the target is to support all that Apache Parquet supports. And one day the current proposal may become hard to maintain

maven-projects/io-api/src/main/java/org/apache/graphar/io/WriteMode.java
```diff
@@ -0,0 +1,28 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.graphar.io;
+
+/** The disposition required for one physical output location. */
+public enum WriteMode {
```
SemyonSinchenko, line 23: Are we going to support `APPEND`?

keksmd:
@SemyonSinchenko Спасибо, поправил все пять пунктов в 5837bd5.

1. `Field` теперь value object (`equals/hashCode` по name/type/nullability).
2. `Schema` сохраняет duplicate names: физический доступ только по индексу.
3. `Row` удалён: `RecordBatch` теперь schema + `ValueVector`-ы + row count, а `VectorRecordBatch` проверяет field order, width и value count.
4. `ColumnType` стал recursive Field-shape: child names/nullability сохраняются для LIST/FIXED_SIZE_LIST/STRUCT/MAP; есть fixed binary и decimal. Так можно расширять Parquet/Arrow shapes, не замораживая плоский enum. Arrow не стал обязательной зависимостью — адаптеры сами владеют native vector lifetime.
5. Добавил `APPEND` и контрактный тест с реальным накоплением значений.

Focused verify: 33 tests, 0 failures; spotless и pre-commit также прошли.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ColumnType.java
```diff
@@ -0,0 +1,208 @@
...
+        return List.copyOf(fields);
+    }
+
+    private static boolean isScalar(Kind kind) {
+        return kind != Kind.FIXED_SIZE_BINARY
+                && kind != Kind.DECIMAL
+                && kind != Kind.LIST
+                && kind != Kind.FIXED_SIZE_LIST
+                && kind != Kind.STRUCT
+                && kind != Kind.MAP;
```
SemyonSinchenko: Any Kind constant added to the enum in the future will be silently treated as a scalar and accepted by of(Kind), bypassing the dedicated-factory guard even if the new kind requires parameters.

I would say we should switch to case + default that throws an exception.

Like:

```java
switch (kind) {
    case BOOLEAN:
    case INT8:
    ...
    case` TIMESTAMP_MILLIS:
        return true;
    case FIXED_SIZE_BINARY:
    ....
    case MAP:
        return false;
    default:
        throw new IllegalStateException("Unhandled kind: " + kind);
}
```

maven-projects/io-api/src/main/java/org/apache/graphar/io/WriteRequest.java
```diff
@@ -0,0 +1,51 @@
...
+ * under the License.
+ */
+
+package org.apache.graphar.io;
+
+import java.net.URI;
+import java.util.Objects;
+
+/** An immutable physical write operation for one schema and output location. */
+public final class WriteRequest {
```
SemyonSinchenko, line 26: I guess we can see in the future something like request deduplication or queue of requests... Let's add `equals` / `hashCode` ? It will cost us almost nothing: uri, schema, mode are already supporting equals and hash.

SemyonSinchenko:
@keksmd Спасибо! Мне нравится, я оставил пару минорных комментов

@keksmd Thanks! Much better, I left two minor comments.

keksmd:
@SemyonSinchenko Спасибо за второй проход — сверил весь набор review-комментариев и поправил оставшиеся два minor пункта в `fa967da`:

1. `ColumnType.isScalar` теперь исчерпывающий `switch`: новый `Kind` нельзя молча принять как scalar; unhandled value падает явно.
2. Добавил value semantics для `Schema` и `WriteRequest` (`equals`/`hashCode`) и тесты на одинаковые независимые schema/request.

Предыдущие пять замечаний из первого review также остаются закрытыми: value `Field`, duplicate schema names, vector `RecordBatch`, recursive Arrow-shaped `ColumnType`, и `APPEND`.

Проверка: `mvn --no-transfer-progress -q -f maven-projects/pom.xml -pl io-api -am verify` — 35 tests, 0 failures; Spotless, pre-commit и `git diff --check` прошли.

Когда будет минутка, пожалуйста, посмотри ещё раз и, если всё ок, approve.

SemyonSinchenko (approved):
~LGTM

SemyonSinchenko:
+ merge conflicts
cc: @keksmd

keksmd:
Rebased onto current `main` to resolve merge conflicts (pom.xml module list). No functional changes — 35 tests still pass. Ready for merge.

SemyonSinchenko:
Nice work!

keksmd:
cc @SemyonSinchenko @SYaoJun — rebased and mergeable now, 35 tests green.

Issue #965, opened 2026-08-24
[FEATURE][Java] Add an S3 storage adapter


keksmd:
### Describe the enhancement requested

Add an optional `graphar-storage-s3` adapter for the dependency-light storage API merged in #958 and #960. It must make GraphAr physical readers and writers able to access `s3://` objects without adding AWS dependencies to metadata or the storage API.

Component(s): Java

## Scope

- Add an optional module depending on `graphar-storage-api` and AWS SDK v2 S3 only.
- Implement validated `s3://bucket/key` URI handling, `HEAD` existence/length, independent seekable inputs, byte-range `GET` reads, and staged local output with conditional create versus overwrite.
- Keep caller ownership of the injected `S3Client`; no global client lifecycle.
- Add credential-free unit coverage with an in-process S3-compatible HTTP endpoint for range, position, EOF, URI validation, and create semantics.

## Non-goals

No GraphAr layout reader/writer, Parquet integration, Iceberg, Hadoop/S3A, Arrow, or mandatory cloud credentials. A real MinIO transport test follows as a separately reviewable opt-in test slice.

## Compatibility and acceptance

The adapter remains Java 11 compatible and optional. Reads must use inclusive S3 Range requests while preserving the `SeekableInput` contract; a fresh open owns an independent position. `create()` must not overwrite an existing object and `createOrOverwrite()` may replace it. Focused Maven verification and pre-commit must pass.

## Links

Relates to storage umbrella #953 and reader/writer umbrella #947. This is deliberately independent of PR #961, which changes the physical batch representation.

keksmd:
Implemented and merged in #966. The module landed with the review changes requested there: typed `headObject` failures instead of a blanket catch, a `FileAlreadyExistsException` when the conditional `PutObject` loses a 412 race, and a 1 MiB read-ahead block so a byte-level reader no longer pays one `GetObject` per `read()`. CI builds and tests the module (`SDK_MODULES` now covers `storage-api,storage-local,io-api,storage-s3`).

PR #966, opened 2026-08-24
feat(java): add an optional S3 storage adapter

keksmd:
### Reason for this PR

Implements #965: an optional, dependency-isolated Java S3 storage adapter over the merged storage API.

- Adds `graphar-storage-s3`, backed by caller-owned AWS SDK v2 `S3Client`.
- Validates `s3://` locations, uses HeadObject/HTTP Range GET for independent seekable reads, and stages sequential output locally before conditional publication.
- Adds credential-free S3 contract coverage plus an opt-in test against a real MinIO endpoint for staged create, non-overwrite, overwrite, and direct-buffer range reads.

The production code does not start, configure, or close an S3 client. The MinIO test is intentionally opt-in: normal SDK CI has no Docker or credential requirement.

### Are these changes tested?

Yes.

```shell
mvn --no-transfer-progress -f maven-projects/pom.xml -pl storage-s3 -am clean verify -Dspotless.check.skip=true
GRAPHAR_MINIO_ENDPOINT=http://localhost:19000 mvn --no-transfer-progress -f maven-projects/pom.xml -pl storage-s3 -am -Dtest=MinioS3StorageIntegrationIT -Dsurefire.failIfNoSpecifiedTests=false test
```

Both pass locally, along with Spotless, pre-commit, and `git diff --check`.

Yes. Java users can construct `S3Storage` with their configured AWS SDK v2 client and access GraphAr `s3://` data through the storage API.

## Checklist

- [x] I have self-reviewed this PR
- [ ] I have run `make cpplint` (if this PR contains C++ changes)
- [x] I have run `pre-commit`
- [x] I have run the relevant tests

### DIFF
--- a/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
+++ b/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
@@ -31,7 +31,9 @@ import org.apache.graphar.storage.PositionOutput;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 
 final class S3OutputFile implements OutputFile {
     private final S3Client client;
@@ -57,18 +59,26 @@ final class S3OutputFile implements OutputFile {
 
     /**
      * Rejects an existing object before any byte is staged, so that a caller learns about the
-     * conflict at open time the way the local adapter does. A head request that cannot answer is
-     * not treated as a conflict: the conditional publication in {@link StagedOutput#close()}
-     * remains the atomic guarantee.
+     * conflict at open time the way the local adapter does. Only an absent object lets the caller
+     * proceed; a head request that fails for any other reason surfaces here rather than after a
+     * full staging pass.
      *
      * @throws FileAlreadyExistsException when the object is already present
+     * @throws IOException when the object cannot be inspected
      */
     private void failIfPresent() throws IOException {
         try {
             client.headObject(
                     HeadObjectRequest.builder().bucket(location.bucket).key(location.key).build());
-        } catch (RuntimeException absentOrUnknown) {
+        } catch (NoSuchKeyException absent) {
             return;
+        } catch (S3Exception exception) {
+            if (exception.statusCode() == 404) {
+                return;
+            }
+            throw new IOException("Cannot inspect S3 object " + location.uri, exception);
+        } catch (RuntimeException exception) {
+            throw new IOException("Cannot inspect S3 object " + location.uri, exception);
         }
         throw new FileAlreadyExistsException(location.uri.toString());
     }
@@ -148,6 +158,11 @@ final class S3OutputFile implements OutputFile {
                     request.ifNoneMatch("*");
                 }
                 client.putObject(request.build(), RequestBody.fromFile(stage));
+            } catch (S3Exception exception) {
+                if (createOnly && exception.statusCode() == 412) {
+                    throw new FileAlreadyExistsException(location.uri.toString());
+                }
+                throw new IOException("Cannot publish S3 object " + location.uri, exception);
             } catch (RuntimeException exception) {
                 throw new IOException("Cannot publish S3 object " + location.uri, exception);
             } finally {
--- a/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
+++ b/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
@@ -23,12 +23,14 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
 import java.nio.ByteBuffer;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import org.apache.graphar.storage.OutputFile;
 import org.apache.graphar.storage.PositionOutput;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
 
 final class S3OutputFile implements OutputFile {
@@ -49,7 +51,26 @@ final class S3OutputFile implements OutputFile {
 
     @Override
     public PositionOutput create() throws IOException {
+        failIfPresent();
         return open(true);
+    }
+
+    /**
+     * Rejects an existing object before any byte is staged, so that a caller learns about the
+     * conflict at open time the way the local adapter does. A head request that cannot answer is
+     * not treated as a conflict: the conditional publication in {@link StagedOutput#close()}
+     * remains the atomic guarantee.
+     *
+     * @throws FileAlreadyExistsException when the object is already present
+     */
+    private void failIfPresent() throws IOException {
+        try {
+            client.headObject(
+                    HeadObjectRequest.builder().bucket(location.bucket).key(location.key).build());
+        } catch (RuntimeException absentOrUnknown) {
+            return;
+        }
+        throw new FileAlreadyExistsException(location.uri.toString());
     }
 
     @Override
--- a/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
+++ b/maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.graphar.storage.s3;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.URI;
+import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.apache.graphar.storage.OutputFile;
+import org.apache.graphar.storage.PositionOutput;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+
+final class S3OutputFile implements OutputFile {
+    private final S3Client client;
+    private final S3Storage.Location location;
+    private final Path stagingDirectory;
+
+    S3OutputFile(S3Client client, S3Storage.Location location, Path stagingDirectory) {
+        this.client = client;
+        this.location = location;
+        this.stagingDirectory = stagingDirectory;
+    }
+
+    @Override
+    public URI uri() {
+        return location.uri;
+    }
+
+    @Override
+    public PositionOutput create() throws IOException {
+        return open(true);
+    }
+
+    @Override
+    public PositionOutput createOrOverwrite() throws IOException {
+        return open(false);
+    }
+
+    private PositionOutput open(boolean createOnly) throws IOException {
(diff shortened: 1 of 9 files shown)


keksmd:
Hi @SemyonSinchenko — this is the next independent Java slice after #960/#962: an optional S3 adapter over the now-merged storage API. It is intentionally isolated from #961 and has credential-free range/write contract coverage. I would appreciate your review when you have time.

keksmd:
Updated this PR into the complete S3 transport vertical: it now includes the opt-in real-MinIO gate (staged conditional publication and range reads), while normal CI stays credential- and Docker-free.

keksmd:
cc @SemyonSinchenko @SYaoJun — green CI, would appreciate a look when you have time.

keksmd:
@SemyonSinchenko @SYaoJun a review here would be appreciated whenever you have time. Rebased on current `main`, no conflicts, CI green.

This follows the same shape as the local adapter merged in #960: it stays behind the storage API and keeps GraphAr layout and format concerns out of storage.

SYaoJun:
> @SemyonSinchenko @SYaoJun a review here would be appreciated whenever you have time. Rebased on current `main`, no conflicts, CI green.
> 
> This follows the same shape as the local adapter merged in #960: it stays behind the storage API and keeps GraphAr layout and format concerns out of storage.

@keksmd Thanks for your contribution!

I am not very familiar with the Java codebase, so it would be helpful to also get a review from Sem or Xiaokang.

I also ran this PR through an AI review tool and received some suggestions. I will paste them below for reference.

I have a few questions about the overall design. I may be missing some context here, but I am trying to understand the motivation behind this PR.

Why does GraphAr need to add S3 storage support? I noticed that there is already a `java-info` directory containing a pure Java reader and writer. Could you explain why the new implementation is not built on top of the existing structure?

Also, could you clarify the relationship between the new S3-related code and the existing `java-info` implementation?

It would be very helpful if you could add a high-level architecture diagram showing the relationships between these components and the overall design rationale.

SYaoJun:
Thanks for the PR — the module is cleanly separated (optional dependency, caller-owned `S3Client`, conditional `PutObject`, eTag/version-pinned reads) and the unit/IT split is well done. A few points should be resolved before this is ready to merge:

**1. The new module is not covered by CI**
`.github/workflows/java-sdk.yml` does not trigger on, nor build, `maven-projects/storage-s3`:
- `paths:` for both `push` and `pull_request` don't include `maven-projects/storage-s3/**`, so the workflow didn't even run for this PR's module.
- `SDK_MODULES: storage-api` means the compile/test jobs never touch `storage-s3` (this PR only stayed green because the root `maven-projects/pom.xml` changed).

Please add the module to the trigger paths and the module list, e.g.:
```yaml
paths:
  - 'maven-projects/storage-s3/**'
...
env:
  SDK_MODULES: storage-api,storage-local,storage-s3
```
(Also note `storage-local` from #960 is currently missing from `SDK_MODULES`, so aligning this list fixes a pre-existing gap too.)

**2. `create()` fails late, unlike the local adapter — inconsistent behavior under the same `Storage` API**
`LocalOutputFile.open()` fails immediately when the file exists (`CREATE_NEW`), but `S3OutputFile.open()` only stages a temp file and defers the existence check until `close()`'s conditional `PutObject` (`If-None-Match: *`). A caller can end up buffering/writing the whole object before hitting the "already exists" error. Suggestion: issue a `headObject` fast-fail in `create()` at open time (covers the common case), and keep `If-None-Match` on publish as the atomic backstop.

**3. Inconsistent license headers**
`S3InputFile.java` and `S3SeekableInput.java` use a single-line `/* ... */` ASF header while every other file in the module (and in `storage-api` / `storage-local`) uses the standard multi-line header. Please align those two files with the canonical header format.

Happy to discuss any of these further. Thanks for the well-structured contribution.

keksmd:
@SYaoJun thanks for the careful read. All three points are addressed in the three commits I just pushed, and the design questions are answered below.

**1. CI coverage — fixed in `ci(java): build the SDK modules that the workflow already watches`**

You are right, and the gap was wider than this module: the workflow already *triggers* on `storage-local/**`, `io-api/**` and `core/**`, but `SDK_MODULES: storage-api` meant none of them were ever compiled or tested upstream. That is my own regression from #962, where `storage-api` genuinely was the only SDK module at the time. The workflow now watches `maven-projects/storage-s3/**` as well and builds `storage-api,storage-local,io-api,storage-s3`.

Verified locally with exactly the module set CI will use, reading the surefire reports rather than the exit code:

```
graphar-storage-api ..... SeekableInputTest 3/0/0
graphar-storage-local ... LocalStorageTest 5/0/0
graphar-io-api .......... 35 tests across 7 classes, 0 failures, 0 errors
graphar-storage-s3 ...... S3StorageTest 2/0/0
```

**2. Late failure in `create()` — fixed in `fix(java): reject an existing S3 object before staging bytes`**

`create()` now heads the object at open time and throws `FileAlreadyExistsException`, the same type the local adapter surfaces from `CREATE_NEW`. The conditional `PutObject` stays as the atomic backstop, so a head request that cannot answer (throttled, denied) does not turn into a spurious conflict — it just falls through to publication. Covered by a new unit test asserting that nothing is published when `create()` is rejected.

**3. License headers — fixed in `style(java): use the canonical ASF header in the S3 input files`**

Both files now carry the standard multi-line header.

---

**On the design questions**

`info` is metadata only: it loads and saves `*.graph.yml` / `*.vertex.yml` / `*.edge.yml` into `GraphInfo`, `VertexInfo`, `EdgeInfo`. Its `src/main/java` tree is `loader`, `saver`, `yaml`, `type` — there is no code there that reads or writes a data chunk, so there is nothing in it for an S3 adapter to build on. The name `ReaderGraphInfoLoader` refers to a `java.io.Reader`, not to a data reader.

The pure-Java SDK is being built as a stack of narrow layers, each merged as its own slice:

```
info          metadata: parse and serialize GraphAr YAML
storage-api   bytes: URI, InputFile/OutputFile, SeekableInput/PositionOutput   (#958)
  storage-local   local filesystem backend                                     (#960)
  storage-s3      S3 backend                                                   (this PR)
io-api        format: batches, schema, read and write requests                 (#961, #971, #972)
core          layout: chunk ranges, offsets, adjacency resolution              (#968)
```

The rule the layers follow is that storage knows only URIs and byte streams — no GraphAr layout, no Parquet, no Hadoop, no Arrow, no query semantics. That is what keeps a backend swappable: `S3Storage` implements the same `Storage` interface as `LocalStorage`, so everything above it is unchanged whether a graph lives on a disk or in a bucket, and the AWS SDK stays an optional dependency that a user who does not need S3 never pulls in.

So this PR does not duplicate anything in `info`; it sits under it, and the reader vertical that will consume both is the layer still to come.

@SemyonSinchenko a second pair of eyes on the Java side would be welcome, as SYaoJun suggested.

keksmd:
Follow-up: widening `SDK_MODULES` made the `Build Java Docs` step fail, and I have fixed that in `ci(java): bind the javadoc step to a lifecycle phase`.

The step invoked the bare goal `mvn -pl ${SDK_MODULES} -am javadoc:javadoc`, so the reactor produced no jars and `graphar-storage-local` tried to resolve `graphar-storage-api:jar:0.13.0-SNAPSHOT` from the repository, where no snapshot is published. With a single module in the list nothing was ever resolved across modules, so the latent problem stayed invisible. Running `package` in the same invocation makes the reactor supply the dependency.

Worth noting for a later cleanup, not changed here: the `verify` step already runs `javadoc:jar (attach-javadocs)` for every module, so `Build Java Docs` duplicates it. Happy to drop the step in a separate PR if you agree.

CI is green now, and the run actually exercised the new modules rather than passing them by:

```
storage-api ...... SeekableInputTest 3/0/0
storage-local .... LocalStorageTest 5/0/0
io-api ........... 35 tests, 0 failures, 0 errors
storage-s3 ....... S3StorageTest 2/0/0
Build Java Docs .. all four modules SUCCESS
```

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
```diff
@@ -0,0 +1,164 @@
...
+     * not treated as a conflict: the conditional publication in {@link StagedOutput#close()}
+     * remains the atomic guarantee.
+     *
+     * @throws FileAlreadyExistsException when the object is already present
+     */
+    private void failIfPresent() throws IOException {
+        try {
+            client.headObject(
+                    HeadObjectRequest.builder().bucket(location.bucket).key(location.key).build());
+        } catch (RuntimeException absentOrUnknown) {
```
SemyonSinchenko: This catch treats every headObject failure (403 auth/permission errors, throttling, network faults) as "object absent", masking real environment errors and silently deferring the true failure to close() with a different message. Compare with S3Storage.exists() in the same module, which correctly distinguishes NoSuchKeyException/404 from other failures. Narrow the catch to the absence case so genuine errors surface at create() time instead of wasting staging work.

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
```diff
@@ -0,0 +1,164 @@
...
+            closed = true;
+            try {
+                output.close();
+                PutObjectRequest.Builder request =
+                        PutObjectRequest.builder().bucket(location.bucket).key(location.key);
+                if (createOnly) {
+                    request.ifNoneMatch("*");
+                }
+                client.putObject(request.build(), RequestBody.fromFile(stage));
+            } catch (RuntimeException exception) {
```
SemyonSinchenko, line 166: When `createOnly` publication loses a concurrent race, S3 rejects the conditional put with 412, which the AWS SDK surfaces as an `S3Exception` — this generic catch wraps it into a plain `IOException`, so callers cannot distinguish "already exists" from a transient failure. This is inconsistent with the eager check in `failIfPresent()` (which throws `FileAlreadyExistsException`) and with the `OutputFile.create()` contract of failing on an existing file. The conditional put is the documented atomic guarantee, so its conflict outcome should surface as the same typed exception. Fully staged data is also dropped with a misleading message in this case.

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3SeekableInput.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        if (position >= size) {
+            return -1;
+        }
+        int count = (int) Math.min(destination.remaining(), size - position);
+        GetObjectRequest.Builder request =
+                GetObjectRequest.builder()
+                        .bucket(location.bucket)
+                        .key(location.key)
+                        .range("bytes=" + position + "-" + (position + count - 1));
```
SemyonSinchenko: Every `read()` issues a full S3 `GetObject` range request with no read-ahead buffering, so byte-level readers (header/footer parsing and small-buffer scans typical of columnar formats, which this `SeekableInput` SPI is designed to serve) pay one network round-trip per read call. Consider buffering ahead (e.g., read a configurable block such as 64KB–1MB into memory and serve subsequent sequential reads from the buffer), so bulk and small reads both amortize to ~1 request per block.

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
keksmd: Narrowed to the absence case as you suggested: only `NoSuchKeyException` and a 404 `S3Exception` mean absent, everything else is now raised as an `IOException` from `create()` rather than deferred to `close()` under a different message. Same shape as `S3Storage.exists()` in this module. New test `surfacesAFailedExistenceCheckInsteadOfStaging` drives a 403 head and asserts nothing was published.

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3OutputFile.java
keksmd: Agreed, the conditional put is the atomic guarantee, so its conflict should be typed. A 412 on a `createOnly` publication now surfaces as `FileAlreadyExistsException` with the object URI, matching `failIfPresent()` and the `OutputFile.create()` contract; any other `S3Exception` keeps the `IOException` wrapping. New test `reportsALostPublicationRaceAsAnExistingObject` covers the raced path.

maven-projects/storage-s3/src/main/java/org/apache/graphar/storage/s3/S3SeekableInput.java
keksmd: Added read-ahead. The input now fetches a 1 MiB block and serves subsequent reads from memory; a read at least as large as the block bypasses the buffer so bulk reads stay at one request and do not pay a copy. New test `servesSequentialReadsFromOneRangeRequest` reads six single bytes and asserts exactly one `GetObject` call, where the old code issued six. The existing range assertion moved from `bytes=2-4` to `bytes=2-5` because the block is now clipped to the object end rather than to the caller buffer.

SemyonSinchenko (approved):
~LGTM! Nice work!

SemyonSinchenko:
Hi @SYaoJun ! LGTM from my point of view. What do you think?

keksmd:
Thanks @SemyonSinchenko for the review and the approval — the three points you raised are what made this module honest: typed failures on `headObject`, a `FileAlreadyExistsException` on the lost 412 race, and read-ahead so a byte-level reader no longer pays a round trip per `read()`.

@SYaoJun your three points are addressed as well (CI now builds `storage-api,storage-local,io-api,storage-s3`, `create()` fast-fails at open time, canonical ASF headers), and the layer diagram in my earlier comment answers the design question about `info`. CI is green on `f1d0bb2` with real test counts: `S3StorageTest` 5/0/0, `io-api` 35/0/0, and `Build Java Docs` now succeeds for all five modules.

Let me know if anything else is needed here.

Issue #967, opened 2026-08-24
[FEATURE][Java] Add chunk and edge-range value primitives


keksmd:
### Describe the enhancement requested

Add the small, storage-free value primitives needed to translate GraphAr ordered-adjacency offsets into exact half-open edge-chunk ranges. This is the first core slice of the pure-Java ordered adjacency reader; metadata URI resolution and physical reading follow separately.

Component(s): Java

## Scope

- Add `graphar-core` with long-safe chunk-index, in-chunk offset, and chunk-count arithmetic.
- Add validated half-open `ChunkRange` and `EdgeRange` values, including correct end-exclusive edge-chunk selection.
- Add unit tests for empty ranges, chunk boundaries, and invalid IDs/sizes/offsets.

## Non-goals

No storage, Parquet, S3, Arrow, metadata URI resolution, GraphAr reader facade, offsets file decoder, or adjacency data read.

## Compatibility and acceptance

The API remains Java 11 compatible and depends on no physical IO module. It must reject negative IDs and non-positive chunk sizes. A range `[begin,end)` selects no chunks when empty; nonempty ranges include the chunk containing `end - 1`, not an extra chunk at `end`. Focused Maven verification and pre-commit must pass.

## Links

Relates to #947, #944, and architecture umbrella #756. The follow-up metadata resolver will use `EdgeInfo` URI methods rather than reconstruct GraphAr paths.

PR #968, opened 2026-08-24
feat(java): add chunk and edge-range primitives

keksmd:
### Reason for this PR

Implements #967 (and relates to #947/#944): the complete dependency-light Java core contract for resolving an ordered GraphAr adjacency range.

- Adds `graphar-core` with long-safe chunk arithmetic and half-open edge/chunk range value types.
- Validates ordered offset chunks, including monotonicity and terminal edge-count checks.
- Resolves `EdgeInfo` ordered-by-source/destination metadata into the exact offset location, edge range, edge-count URI, and selected adjacency chunk URIs.
- Verifies the canonical LDBC metadata layout and the cross-edge-chunk boundary `[1008, 1061)` at chunk size `1024`.

This PR delegates URI construction to `EdgeInfo`; it deliberately does not read Parquet offsets or depend on storage/Parquet. That physical reader belongs after the open #961 IO API change.

### Are these changes tested?

Yes.

```shell
GAR_TEST_DATA=/private/tmp/graphar-pr967/testing mvn --no-transfer-progress -f maven-projects/pom.xml -pl core -am clean verify -Dspotless.check.skip=true
mvn --no-transfer-progress -f maven-projects/pom.xml -pl core -am spotless:check
```

The reactor passes 8 core tests plus the `graphar-info` dependency suite. `pre-commit run --files ...` and `git diff --check` also pass.

Yes. Java consumers can now resolve ordered-adjacency metadata and validated offset ranges to exact GraphAr chunk URIs without introducing physical IO dependencies.

## Checklist

- [x] I have self-reviewed this PR
- [ ] I have run `make cpplint` (if this PR contains C++ changes)
- [x] I have run `pre-commit`
- [x] I have run the relevant tests

### DIFF
--- a/maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
+++ b/maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.graphar.core;
+
+/** A half-open range of non-negative chunk indexes. */
+public final class ChunkRange {
+    private final long begin;
+    private final long end;
+
+    /**
+     * Creates a chunk range.
+     *
+     * @param begin the first included non-negative chunk index
+     * @param end the first excluded chunk index
+     */
+    public ChunkRange(long begin, long end) {
+        if (begin < 0) {
+            throw new IllegalArgumentException("Chunk range begin must be non-negative: " + begin);
+        }
+        if (end < begin) {
+            throw new IllegalArgumentException(
+                    "Chunk range end must not precede begin: [" + begin + ", " + end + ")");
+        }
+        this.begin = begin;
+        this.end = end;
+    }
+
+    /**
+     * Returns the first included chunk index.
+     *
+     * @return the first included chunk index
+     */
+    public long begin() {
+        return begin;
+    }
+
+    /**
+     * Returns the first excluded chunk index.
+     *
+     * @return the first excluded chunk index
+     */
+    public long end() {
+        return end;
+    }
+
+    /**
+     * Returns whether this range selects no chunks.
+     *
+     * @return whether the range is empty
+     */
+    public boolean isEmpty() {
+        return begin == end;
+    }
+
+    /**
+     * Returns whether {@code chunkIndex} belongs to this range.
+     *
+     * @param chunkIndex a chunk index
+     * @return whether the chunk index belongs to this range
+     */
+    public boolean contains(long chunkIndex) {
+        return chunkIndex >= begin && chunkIndex < end;
+    }
+}
--- a/maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
+++ b/maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
@@ -19,6 +19,8 @@
 
 package org.apache.graphar.core;
 
+import java.util.Objects;
+
 /** A half-open range of non-negative chunk indexes. */
 public final class ChunkRange {
     private final long begin;
@@ -77,5 +79,27 @@ public final class ChunkRange {
      */
     public boolean contains(long chunkIndex) {
         return chunkIndex >= begin && chunkIndex < end;
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (!(other instanceof ChunkRange)) {
+            return false;
+        }
+        ChunkRange that = (ChunkRange) other;
+        return begin == that.begin && end == that.end;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(begin, end);
+    }
+
+    @Override
+    public String toString() {
+        return "ChunkRange[" + begin + ", " + end + ")";
     }
 }
--- a/maven-projects/core/src/main/java/org/apache/graphar/core/EdgeRange.java
+++ b/maven-projects/core/src/main/java/org/apache/graphar/core/EdgeRange.java
@@ -19,6 +19,8 @@
 
 package org.apache.graphar.core;
 
+import java.util.Objects;
+
 /** A validated half-open range of edge rows within one GraphAr vertex partition. */
 public final class EdgeRange {
     private final long begin;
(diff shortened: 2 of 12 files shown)


keksmd:
Hi @SemyonSinchenko — this is a deliberately small, independent Java core slice (419 lines, 6 focused tests) after #960/#962. It provides only the long-safe half-open chunk/range value layer; metadata URI resolution and IO stay out of this PR. I would appreciate your review when convenient.

keksmd:
Updated this PR into the complete ordered-adjacency core vertical (now 848 additions): validated offset chunks, metadata URI resolution, and the canonical LDBC cross-chunk fixture. It remains independent of #961 by keeping Parquet reads out of core.

keksmd:
cc @SemyonSinchenko @SYaoJun — green CI, would appreciate a look when you have time.

keksmd:
@SemyonSinchenko @SYaoJun could you take a look when you have a moment? This one is rebased on current `main`, conflicts are resolved and CI is green.

Note that the `test` job only builds `-pl storage-api -am`, so `graphar-core` is not exercised upstream yet. Locally the module is green: `OrderedAdjacencyResolverFixtureTest` 2/0/0 and `ChunkRangeTest` 6/0/0 from `core/target/surefire-reports`. The fixture test resolves the canonical data through `GAR_TEST_DATA` (then `gar.test.data`, then the usual relative fallbacks) and skips cleanly when the testing repository is absent, matching the convention in `graphar-info`.

I can follow up with a separate PR widening `SDK_MODULES` so the newer SDK modules are actually compiled and tested in CI, if that sounds right to you.

keksmd:
Update on the CI gap I mentioned above: the fix landed in #966 rather than as a separate PR, since @SYaoJun raised the same point there. The workflow now builds `storage-api,storage-local,io-api,storage-s3`.

`core` is deliberately not in that list yet — it only exists once this PR merges. I will add it here, or in a follow-up, whichever you prefer.

keksmd:
Rebased on `main` after #966 merged — the only conflict was the module list in `maven-projects/pom.xml`, where `storage-s3` and `core` were added at the same line; both are kept.

I also pushed the answer to the CI question I raised earlier, rather than leaving it open: the workflow already watches `maven-projects/core/**` in its path filters, so a change here triggers the job, but `SDK_MODULES` left `core` out and the reactor never compiled or tested it — a green run proved nothing about this module. `SDK_MODULES` is now `storage-api,storage-local,io-api,storage-s3,core`. The fixture-backed test skips itself when the canonical `testing` dataset is absent, so the job stays green without the submodule and gains real coverage when it is present.

Local verification against `main` with the fixtures available:

```
ChunkRangeTest                        Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
OrderedAdjacencyResolverFixtureTest   Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

maven-projects/core/src/main/java/org/apache/graphar/core/OrderedAdjacencyResolver.java
```diff
@@ -0,0 +1,83 @@
...
+    public ResolvedAdjacency resolve(long vertexId, long offsetBegin, long offsetEnd) {
+        OffsetLocation offsetLocation = locate(vertexId);
+        EdgeRange edgeRange = EdgeRange.fromOffsets(offsetBegin, offsetEnd);
+        return resolved(offsetLocation, edgeRange);
+    }
+
+    /** Resolves a vertex using a complete, validated offset chunk read by a physical backend. */
+    public ResolvedAdjacency resolve(long vertexId, OffsetChunk offsetChunk) {
+        OffsetLocation offsetLocation = locate(vertexId);
+        Objects.requireNonNull(offsetChunk, "Offset chunk cannot be null.");
```
SemyonSinchenko, line 70: Two robustness gaps in this overload: (1) `Objects.requireNonNull(offsetChunk)` runs after `locate(vertexId)`, so when both arguments are invalid the null-chunk error is masked by the vertexId error — validate arguments up front. (2) `OffsetChunk` carries no identity of the vertex chunk it was read from, so passing a chunk read via a different `locate()` call (e.g., a cached or stale chunk) is silently accepted and yields a plausible but wrong edge range. Consider recording the source vertex chunk index in `OffsetChunk` (factory like `of(long vertexChunkIndex, long[] offsets)`) and verifying it matches `offsetLocation.vertexChunkIndex()` here, turning silent misuse into a clear error.

maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
SemyonSinchenko, line 1: This composite value type (like ChunkRange, EdgeRange, and OffsetLocation) lacks equals/hashCode/toString. Without them, callers cannot assert whole results in tests (current tests must compare field by field), use results in collections, or log a diagnostic summary of a resolved adjacency. Consider implementing equals/hashCode over the final fields and a toString that summarizes the vertex chunk, edge range, and chunk range.

maven-projects/core/src/main/java/org/apache/graphar/core/EdgeRange.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Returns the half-open range of edge chunks intersecting this edge range.
+     *
+     * @param edgeChunkSize a positive number of edge rows per chunk
+     * @return the chunk range intersecting this edge range
+     */
+    public ChunkRange edgeChunks(long edgeChunkSize) {
+        ChunkMath.validateChunkSize(edgeChunkSize);
```
SemyonSinchenko: This method re-implements, inline, the same arithmetic that ChunkMath centralizes in this package: `begin / edgeChunkSize` is `ChunkMath.chunkIndex(begin, edgeChunkSize)` and `1 + (end - 1) / edgeChunkSize` is exactly `ChunkMath.chunkCount(end, edgeChunkSize)` (the existing tests assert both with identical inputs, e.g. 201/100 -> 3 and Long.MAX_VALUE/1 -> MAX_VALUE). Duplicating the long-safe formulas risks them drifting apart if the overflow-safe expressions in ChunkMath are ever revised. Delegate to ChunkMath so this package has a single home for chunk arithmetic.

SemyonSinchenko:
I left a few minor comments. ~LGTM overall

maven-projects/core/src/main/java/org/apache/graphar/core/EdgeRange.java
keksmd: Done in 2412d03 — `EdgeRange.edgeChunks` now calls `ChunkMath.chunkIndex(begin, size)` and `ChunkMath.chunkCount(end, size)` instead of repeating the division, so the arithmetic (and its validation) lives in one place.

maven-projects/core/src/main/java/org/apache/graphar/core/ChunkRange.java
keksmd: Added in 2412d03: `equals`/`hashCode`/`toString` on `ChunkRange`, `EdgeRange` and `OffsetLocation`. The fixture test now asserts whole values (`assertEquals(new ChunkRange(0, 2), resolved.edgeChunks())`) instead of begin/end field by field. `ResolvedAdjacency` got `toString` only — it holds an `EdgeInfo`, which has no value equality, so equals there would not be well defined.

maven-projects/core/src/main/java/org/apache/graphar/core/OrderedAdjacencyResolver.java
keksmd: Fixed in 2412d03, and the null check was only half the problem. `OffsetChunk.of` now takes the vertex chunk index it was read from and exposes `vertexChunkIndex()`; `resolve(vertexId, offsetChunk)` does `requireNonNull` first, then rejects a chunk whose index does not match the located vertex chunk. Before this, passing an offset chunk from another vertex chunk returned a plausible but wrong edge range instead of failing. New test `rejectsAnOffsetChunkReadFromAnotherVertexChunk` pins it. The API is unreleased, so I changed the factory rather than adding an overload.

SemyonSinchenko (approved):

PR #971, opened 2026-09-01
feat(java): let a read request name what it wants pushed down

keksmd:
Adds the request-side value types for a pushdown-capable read: `Projection`, `RowRange`, `Filter`, `ComparisonOperator`, `Literal`, `ReadRequest`, and the `ReadCapability` enum a physical reader declares support for.

This is the next slice after #961 (merged) — it only describes what a caller can ask a reader to push down. It does not add a reader, a result type, or IO. That's the follow-up PR.

Focused verify: `mvn --no-transfer-progress -f maven-projects/pom.xml -pl io-api -am spotless:check` and `clean verify -Dspotless.check.skip=true` — clean, tests green.


keksmd:
cc @SemyonSinchenko @SYaoJun — next slice after #961 (merged). Green CI.

keksmd:
@SemyonSinchenko @SYaoJun this is the read half of the IO API you reviewed in #961, so it may be easiest to look at with that context. Rebased on current `main`, no conflicts, CI green.

It adds the request vocabulary only: projection, optional row range, optional predicate and limit. No query AST and no Arrow dependency in the public surface.

keksmd:
Rebased onto current `main` (now including #966 and #968) and force-pushed, so CI here runs against what would actually merge rather than the older base. Locally `mvn -pl io-api -am clean verify` gives 53 tests, 0 failures, 0 errors, 0 skipped.

keksmd:
@SemyonSinchenko when you have a moment — this one is green and rebased on current `main`. It is the smaller half of the read contract: the request object plus the pushdown vocabulary, no behaviour yet.

I cannot add reviewers on this repo, so a ping is the only lever I have. Merging this before #972 keeps that one down to a single commit.

keksmd:
Superseded: #972 was merged with these commits stacked underneath it, so the request object and pushdown vocabulary are already on main (6f97f01). Closing to avoid a duplicate.

Issue #964, opened 2026-08-23
[FEATURE][Java] Adapt Parquet physical IO to columnar batches


keksmd:
### Describe the enhancement requested

After #959 / PR #961 replaces the provisional row-oriented physical-batch contract with
column vectors, the Parquet backend needs the corresponding adapter migration. This issue is one
narrow downstream slice: it must not retain a compatibility `Row` facade.

Scope:

- migrate `graphar-io-api` read-contract tests and `graphar-io-parquet` from
  `RecordBatch.row(int)` / `Row` to `ValueVector` and `VectorRecordBatch`;
- preserve physical projection, strict half-open row-range reads, and required Parquet Offset Index
  behavior;
- preserve safe binary/list/null value ownership at the vector boundary;
- treat `WriteMode.APPEND` explicitly: Parquet cannot atomically append to one existing file, so the
  writer must reject it rather than silently overwriting.

Non-goals: migrating GraphAr reader/writer/validator/integration consumers (separate dependent
slices); adding Apache Arrow as a dependency; changing the Parquet decoding algorithm; or making
Parquet append work through a hidden rewrite.

Acceptance evidence:

1. no `Row` or `.row(...)` reference remains in `io-api` / `io-parquet`;
2. focused clean verification covers vector schema position, count, null, binary and list values;
3. canonical fixture projection/row-range tests retain applied `PROJECTION` and `ROW_RANGE`;
4. writer still emits readable Offset Indexes with the existing bounded-page assertion;
5. a write with `APPEND` is explicitly rejected, never treated as overwrite.

Component(s): Java

Relates to #947 and #959.

keksmd:
Status and the exact gate on this slice.

The module is ready in migrated form — no \`Row\` and no \`.row(...)\` anywhere in it; batches are \`ValueVector\` / \`VectorRecordBatch\`, and \`ParquetRow\` is replaced by \`ParquetValueVector\`. I verified it compiles and tests clean against `main` plus the read contract from #972:

```
graphar-io-parquet ................................. SUCCESS
ParquetPhysicalReaderFixtureTest      Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
ParquetPhysicalWriterOffsetIndexTest  Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
ParquetPhysicalIoEfficiencyTest       Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
[...]

PR #972, opened 2026-09-01
feat(java): make a declined pushdown visible instead of silent

keksmd:
Adds the reader side: `PhysicalReader`, `ReadReport`, `ReadResult`, plus a contract test that pins the invariant a physical reader must hold — a reader that declines every pushdown hint returns the same rows and the same schema as one that applies all of them. Only the reported `applied`/`declined` sets differ.

Builds on #971 (this PR's base branch). Together with the merged #961 (batch/write shapes) and #971 (request shapes), this completes the read half of the IO API.

Focused verify: `mvn --no-transfer-progress -f maven-projects/pom.xml -pl io-api -am spotless:check` and `clean verify -Dspotless.check.skip=true` — clean, 63 tests green.


keksmd:
cc @SemyonSinchenko @SYaoJun — stacked on #971 (shows its commits until that merges, diff will shrink automatically). Green CI.

keksmd:
@SemyonSinchenko @SYaoJun this one builds on #971, so please review that first. It is opened against `main` because GitHub cannot base a pull request on a branch that lives in a fork, so the diff currently contains the #971 commits as well; it will shrink to its own slice once #971 lands.

The slice itself makes a declined pushdown observable: a backend reports which hints it applied, and declining one must not change the results.

keksmd:
Rebased onto current `main` (now including #966 and #968) and force-pushed. Locally `mvn -pl io-api -am clean verify` gives 63 tests, 0 failures, 0 errors, 0 skipped.

This one sits on top of #971 — reviewing and merging #971 first leaves a single commit here. Together the two are the last piece the Parquet backend in #964 needs, so I am holding that PR until both land rather than opening a dependent chain.

keksmd:
@SemyonSinchenko this is the other half of the read contract — a backend reports which hints it actually applied, so a declined pushdown is visible rather than silently changing nothing.

Green and rebased on current `main`. Best read after #971, which it sits on top of.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+    private final Literal value;
+
+    private Filter(String column, ComparisonOperator operator, Literal value) {
+        if (column == null || column.isBlank()) {
+            throw new IllegalArgumentException("A filter column cannot be blank.");
+        }
+        this.column = column;
+        this.operator = Objects.requireNonNull(operator, "Filter operator cannot be null.");
+        if ((operator == ComparisonOperator.IS_NULL || operator == ComparisonOperator.IS_NOT_NULL)
+                && value != null) {
```
SemyonSinchenko: Should we have `value` as `Option` instead of an explicit `null` contract? Passing `null` for optional value sounds like something from 2010th for me :D

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+    /** Creates a null check for {@code column}. */
+    public static Filter isNull(String column) {
+        return new Filter(column, ComparisonOperator.IS_NULL, null);
+    }
+
+    /** Creates a non-null check for {@code column}. */
+    public static Filter isNotNull(String column) {
+        return new Filter(column, ComparisonOperator.IS_NOT_NULL, null);
+    }
+
```
SemyonSinchenko, line 178: Why not add `Filter.gt`, `Filter.get`, `Filter.neq`, `Filter.lt`, `Filter.leq`, etc. and make the `public static Filter comparison(String column, ComparisonOperator operator, Literal value)` private internal contract?

SemyonSinchenko: In that case we can keep `null` contract and whatever we want, while users will get a clear and intuitive API...

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+ * compared by this contract.
+ *
+ * <p>Comparison operands must have the same declared type; a type mismatch is invalid rather than a
+ * coercion. Null values never match a comparison, including NOT_EQUAL; use IS_NULL or IS_NOT_NULL
+ * for null tests. Ordered STRING comparisons use {@link String#compareTo(String)}, and DATE and
+ * TIMESTAMP_MILLIS use their natural ordering. Readers must reject an invalid comparison before
+ * returning a result, whether the filter is pushed down or evaluated as fallback.
+ */
+public final class Filter {
+    private final String column;
```
SemyonSinchenko: I like the idea of having `Literal` instead of `Object`, but the very logical next step will be to have `Column` instead of `String column`: by the end, filter can be a complex one, can't it?

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+ * uses LocalDate; and TIMESTAMP_MILLIS uses millisecond-precise Instant. Other types cannot be
+ * compared by this contract.
+ *
+ * <p>Comparison operands must have the same declared type; a type mismatch is invalid rather than a
+ * coercion. Null values never match a comparison, including NOT_EQUAL; use IS_NULL or IS_NOT_NULL
+ * for null tests. Ordered STRING comparisons use {@link String#compareTo(String)}, and DATE and
+ * TIMESTAMP_MILLIS use their natural ordering. Readers must reject an invalid comparison before
+ * returning a result, whether the filter is pushed down or evaluated as fallback.
+ */
+public final class Filter {
```
SemyonSinchenko: Should we make compositions possible, like 

```java
var filter1 = Filter.isNotNull("a);
var filter2 = Filter.comparison("a", ComparisonOperator.GREATER_THAN, Literal.of(0.0));

var filter = filter1 && filter2;
```

??

maven-projects/io-api/src/main/java/org/apache/graphar/io/Projection.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+/** An ordered set of requested output columns. */
+public final class Projection {
+    private static final Projection ALL_COLUMNS = new Projection(true, List.of());
+
+    private final boolean allColumns;
+    private final List<String> columns;
```
SemyonSinchenko: I still do not like the idea of working with `String column`. I see a lot of potential risks, like unresolved vs resolved, unqualified vs qualified, etc. etc. As well possible collisions by unqualified names as an example.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ReadRequest.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.OptionalLong;
+import java.util.Set;
+
+/**
+ * An immutable physical read operation. Filters are combined with logical AND and may require a
+ * backend to read columns that are not in the requested output projection.
+ */
+public final class ReadRequest {
```
SemyonSinchenko, line 36: Let's add `equals` / `hashCode`, so the physical execution layer can, for example, deduplicate requests / cache results per request / etc.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ReadReport.java
```diff
@@ -0,0 +1,61 @@
...
+
+package org.apache.graphar.io;
+
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Objects;
+import java.util.Set;
+
+/** Per-read accounting of physical hints applied or declined by a backend. */
+public final class ReadReport {
```
SemyonSinchenko, line 28: I would add `equals` / `hashCode`, so the system can, for example, separate processed / unprocessed and compare them elementwise or use `Set`

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
keksmd: Done: `Filter.Comparison.value()` returns `Optional<Literal>`, absent for null checks. The constructor is private and each factory builds exactly the shape it names, so there is no null contract left on the public surface.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
keksmd: Done: `Filter.equal / notEqual / lessThan / lessThanOrEqual / greaterThan / greaterThanOrEqual / isNull / isNotNull`. `comparison(...)` is private now.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
keksmd: Done: `ColumnRef` replaces `String column` in both `Filter` and `Projection`, and `Filter` is now a predicate tree (see the thread on composition). I went with `ColumnRef` rather than `Column` because `RecordBatch.column(int)` already returns a vector and the two would read as the same thing.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Filter.java
keksmd: Done. `Filter` is a closed tree: `Comparison`, `And`, `Or`, `Not`. Static `Filter.and(a, b, ...)`, `Filter.or(...)`, `Filter.not(x)` plus instance `a.and(b)`, `a.or(b)`, `a.negate()`, so your example is `Filter.isNotNull(a).and(Filter.greaterThan(a, Literal.of(0.0)))`. Same-kind junctions flatten and a double negation collapses, so structurally equal predicates compare equal. A reader walks the tree with `Filter.Visitor<R>` - one callback per node kind - so a new node kind is a compile error in every backend rather than a silently skipped branch. `ReadRequest.filters()` stays a list; it is still the top-level conjunction so a backend can push some terms down and evaluate the rest.

maven-projects/io-api/src/main/java/org/apache/graphar/io/Projection.java
keksmd: Agreed, and this round moves the boundary to a typed `ColumnRef`. On the specific risks: a request is built before its file is opened, so a reference is unresolved by construction; the only place it binds is `Schema.resolve(ColumnRef) -> int` in the backend. That resolution is exact - verbatim name match, unknown column is an error, and a name that matches more than one field is an error (`Schema` allows duplicates, as you noted on #961, so first-match would be silent data corruption). There is no qualifier on purpose: the physical boundary reads one file with one schema, and mapping a qualified logical property onto a physical column is the job of the layer above. Nested paths are not in `ColumnType` yet, so `ColumnRef` is one segment today and grows to a path without breaking callers when a struct kind lands.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ReadRequest.java
keksmd: Done: `ReadRequest.equals/hashCode` over uri, projection, row range, filters and limit, with a test that flips each one.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ReadReport.java
keksmd: Done: `ReadReport.equals/hashCode` over applied and declined sets.

keksmd:
Addressed every thread above. Because the files live in the #971 commit this PR stacks on, the changes landed there (`ColumnRef`, `Filter` as a predicate tree with `Visitor`, named factories, `Optional` value, `Schema.resolve`, `ReadRequest.equals/hashCode`) and this PR was rebased on top with `ReadReport.equals/hashCode`. Locally `mvn -pl io-api -am clean verify`: #971 63 tests, #972 74 tests, 0 failures.

SemyonSinchenko (approved):
~LGTM!

SemyonSinchenko:
Merging this because there are no risks: Java module is not published yet.

PR #975, opened 2026-09-16
feat(java): read and write Parquet through column vectors

### LINKED ISSUE #964: [FEATURE][Java] Adapt Parquet physical IO to columnar batches

keksmd:
### Reason for this PR

Closes #964. #961 replaced the row-oriented physical batch contract with `ValueVector` / `VectorRecordBatch`; this is the Parquet backend for that contract. It is the physical IO layer the pure-Java reader and writer (#947) sit on, and it keeps the layout-free boundary: no GraphAr semantics, no Arrow, no `Row` facade.

New module `graphar-io-parquet` (wired into `maven-projects/pom.xml` and the Java SDK workflow):

- `ParquetPhysicalReader` / `ParquetBatchCursor`: reads one row group per batch, applies physical projection and the strict half-open row range, and reports `PROJECTION` / `ROW_RANGE` as applied. Each batch is a `VectorRecordBatch` of `ParquetValueVector` columns.
- `ParquetValueVector`: binary values come back as read-only `ByteBuffer`s over a private copy and lists as immutable `List`s, so a consumer never aliases the decoder's buffers.
- `ParquetPhysicalWriter`: emits Offset Indexes with bounded pages. `WriteMode.APPEND` is rejected with `UnsupportedOperationException` because Parquet cannot append to a closed file; it is never treated as an overwrite.
- `io-api`: `ObjectValueVector` (in-memory vector) and `RecordBatches.ofRows` / `ofArrays` (row-to-column transposer) so producers can build column batches without Arrow.

Non-goals kept out: GraphAr reader/writer/validator consumers, Arrow, changes to the decoding algorithm.

### Are these changes tested?

Yes, from `maven-projects/` with the same commands the workflow runs:

```
mvn --no-transfer-progress -pl storage-api,storage-local,io-api,io-parquet,storage-s3,core -am spotless:check
mvn --no-transfer-progress -pl storage-api,storage-local,io-api,io-parquet,storage-s3,core -am clean verify -Dspotless.check.skip=true
mvn --no-transfer-progress -pl io-api,io-parquet -am package javadoc:javadoc -DskipTests -Dspotless.check.skip=true
```

All green: io-api 79 tests, io-parquet 8 tests.

- `RecordBatchesTest`: schema position, row/column counts, nulls, width mismatch rejection, list snapshot on read.
- `ParquetPhysicalReaderFixtureTest`: canonical `ldbc_sample` fixture; projection and row-range reads retain applied `PROJECTION` and `ROW_RANGE`.
- `ParquetListRoundTripTest`: nullable string lists round-trip, empty and null preserved.
- `ParquetPhysicalWriterOffsetIndexTest`: written files carry readable Offset Indexes with the bounded-page assertion.
- `ParquetPhysicalWriterModeTest`: `APPEND` throws and leaves the existing file byte-identical; `OVERWRITE` replaces it.
- `ParquetPhysicalIoEfficiencyTest`: projection and row range only touch the needed column chunks / row groups.

`pre-commit run --files` on the changed paths: passed.

New public module `org.apache.graphar:graphar-io-parquet` and two new public io-api classes (`ObjectValueVector`, `RecordBatches`). No changes to existing public APIs.


SemyonSinchenko:
Did we even consider using anything else for parquet reading/writing except the hadoop ecosystem? For example, https://github.com/hardwood-hq/hardwood ? That is a question I would like to discuss first. The problem of hadoop tech stack for me is it is a) heavy b) leads to the dependency hell for consumers

SemyonSinchenko:
@SYaoJun @Thespica @yecol @keksmd @yangxk1 wdyt about the question above (hadoop vs something-else)?

keksmd:
Yes, considered, and I share the concern about the Hadoop tail. Current state and reasoning:

**What this PR actually depends on.** Our code has zero `org.apache.hadoop` imports. GraphAr storage (`storage-api`: `InputFile`/`SeekableInput`/`PositionOutput`) is what Parquet reads from and writes to, via `ParquetInputFile`/`ParquetOutputFile` adapters — no S3A, no Hadoop `FileSystem`. `hadoop-common` and `hadoop-mapreduce-client-core` are there only because `parquet-hadoop` needs `Configuration` and MapReduce input classes on the classpath at runtime. So the Hadoop dependency is an accident of `parquet-java` packaging, not a design choice.

**Why `parquet-java` for the first slice.**
- Java 11 baseline of the SDK. Hardwood requires Java 21 (`maven.compiler.release=21` in its pom; native build on 25).
- Write to object storage: Hardwood's writer targets `OutputFile.of(Path)` only, object-store output is "coming soon" (README, 1.1.0.Beta1, 2026-08-31). Our writer goes through `storage-api`, S3 included, today.
- Maturity as a format oracle: the C++ implementation is the reference, `parquet-java` is the canonical Java counterpart. For a first columnar slice I preferred boring-and-correct.

**What is already backend-neutral.** `io-api` (`PhysicalReader`/`PhysicalWriter`, `RecordBatch`/`ValueVector`, `ReadRequest` with projection/range/filter/limit and the applied-hints report) has no Parquet or Hadoop types. `io-parquet` is one implementation behind it; `reader`/`writer` never see `parquet-java`. Swapping or adding an engine is a module, not an API change.

**What I would propose.**
1. Keep this PR as the `parquet-java` backend, but I can shrink the tail right here: replace `hadoop-common` + `hadoop-mapreduce-client-core` with the shaded `hadoop-client-api` (single jar, no transitive Guava/Jackson/etc.). If you want that in this PR, say so and I will do it.
2. Follow-up issue: a Hardwood backend spike as a separate optional module, Java 21, behind the same `PhysicalReader`/`PhysicalWriter`, with a JMH comparison on topology (INT64 src/dst) and property chunks. Hardwood's typed `ColumnReader`/`ColumnBatch` maps well onto column vectors and would let us add primitive accessors to `ValueVector` without boxing; I would rather do that with numbers than by assumption.
3. Decision needed from maintainers: is Java 21 acceptable for an optional module while the core SDK stays on 11? If yes, Hardwood is a real candidate for the default once its object-store write lands. If no, `parquet-java` + shaded Hadoop client is the realistic option.

Happy to open the follow-up issue if this direction works for you.

SemyonSinchenko:
@keksmd Thanks! Sounds to me. I will try to review the PR ASAP

maven-projects/io-api/src/main/java/org/apache/graphar/io/ObjectValueVector.java
```diff
@@ -0,0 +1,66 @@
...
+public final class ObjectValueVector implements ValueVector {
+    private final Field field;
+    private final List<Object> values;
+
+    public ObjectValueVector(Field field, List<?> values) {
+        this.field = Objects.requireNonNull(field, "A vector field cannot be null.");
+        Objects.requireNonNull(values, "Vector values cannot be null.");
+        List<Object> copy = new ArrayList<>(values.size());
+        for (Object value : values) {
+            copy.add(value instanceof List ? List.copyOf((List<?>) value) : value);
```
SemyonSinchenko: `List.copyOf` throws a message-less `NullPointerException` when a list value contains null elements, yet the io-api type model explicitly permits such values: `ColumnType.listOf(elementType)` builds a list whose element field is nullable, and `listOfElement` preserves element nullability. A producer assembling a batch for a nullable-element list column that contains a null entry therefore cannot construct this vector at all, and fails with an opaque NPE that points neither at the column nor the offending value. Use a null-tolerant unmodifiable copy (e.g., `Collections.unmodifiableList(new ArrayList<>(...)))` so the vector can represent every value its Field's type allows.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ObjectValueVector.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public final class ObjectValueVector implements ValueVector {
+    private final Field field;
+    private final List<Object> values;
+
+    public ObjectValueVector(Field field, List<?> values) {
+        this.field = Objects.requireNonNull(field, "A vector field cannot be null.");
+        Objects.requireNonNull(values, "Vector values cannot be null.");
+        List<Object> copy = new ArrayList<>(values.size());
+        for (Object value : values) {
```
SemyonSinchenko: The snapshot is only one level deep. For nested list values (`ColumnType.listOf(listOf(...)`) is a representable type), the inner lists are neither copied nor made unmodifiable, so the producer's original inner lists remain live references inside the vector and can be mutated after construction. This contradicts the class javadoc's guarantee that neither the producer nor a consumer can mutate the vector afterwards. Either snapshot list values recursively, or narrow the javadoc to state that only the top level of list values is protected.

maven-projects/io-api/src/main/java/org/apache/graphar/io/RecordBatches.java
```diff
@@ -0,0 +1,81 @@
...
+    private RecordBatches() {}
+
+    /**
+     * Transposes row-major values into a columnar batch.
+     *
+     * @param schema the batch schema; every row must hold one value per field
+     * @param rows the rows, each a list of boxed values in schema order
+     * @return a batch backed by {@link ObjectValueVector} columns
+     */
+    public static RecordBatch ofRows(Schema schema, List<? extends List<?>> rows) {
```
SemyonSinchenko: `ofRows` validates only the row width; neither value types nor nullability are checked against the Field definitions. A String in an INT64 column, or a null in a non-nullable column, passes construction silently and only fails later deep inside a format encoder (e.g., `ParquetPhysicalWriter` 's "Unexpected value for ..." / "Required field is null"), far from the site where the bad row was supplied. Since `Field` exposes `type()` and `nullable()`, failing fast here — at minimum rejecting nulls in non-nullable columns — would make producer errors immediately attributable. If deferred validation is intentional, document it on the factory.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalReader.java
```diff
@@ -0,0 +1,426 @@
...
+            throw new IllegalArgumentException("Unsupported Parquet LIST field: " + type);
+        }
+        Type repeated = type.asGroupType().getType(0);
+        if (!repeated.isRepetition(Type.Repetition.REPEATED)
+                || repeated.isPrimitive()
+                || repeated.asGroupType().getFieldCount() != 1) {
+            throw new IllegalArgumentException("Unsupported Parquet LIST field: " + type);
+        }
+        Type element = repeated.asGroupType().getType(0);
+        if (!element.isPrimitive() || !element.isRepetition(Type.Repetition.REPEATED)) {
```
SemyonSinchenko: This requires the LIST element to be REPEATED, but the Parquet spec's 3-level LIST layout defines the element as REQUIRED or OPTIONAL. Files written by Spark/Arrow/pyarrow are therefore rejected here at schema-mapping time, contradicting the "Only primitive or standard LIST Parquet columns are supported" message in columns(). Note that ParquetBatchCursor.listValue already reads values through the middle repeated "list" wrapper group, so accepting spec-conformant elements mainly needs this check relaxed plus null-element handling (element repetition count 0). Also note this check and `ParquetPhysicalWriter.listElementType` must be changed together: the writer currently emits a REPEATED element, so today the module only round-trips its own non-standard layout.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalWriter.java
```diff
@@ -0,0 +1,288 @@
...
+        Type value = listElementType(element);
+        return (field.nullable() ? Types.optionalList() : Types.requiredList())
+                .element(value)
+                .named(field.name());
+    }
+
+    private static Type listElementType(ColumnType element) {
+        LogicalTypeAnnotation logical = logicalType(element);
+        if (logical == null) {
+            return Types.repeated(physicalType(element)).named("element");
```
SemyonSinchenko: See the comment in Reader about "repeated"

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
```diff
@@ -0,0 +1,350 @@
...
+            } finally {
+                fileReader.close();
+            }
+        }
+    }
+
+    private boolean openNextRowGroup() throws IOException {
+        BlockRange rowGroup = nextRange();
+        if (rowGroup == null) return false;
+        pages =
```
SemyonSinchenko: Row-group reads always go through `readFilteredRowGroup` with a RowRanges object, which requires offset indexes in the file. When no row range was requested (rangeStart=0, rangeEnd=MAX_VALUE) or the range covers the whole group, this still forces the offset-index code path, so a plain full scan of a Parquet file without offset indexes (older or foreign writers) throws `MissingOffsetIndexException` and surfaces as `UnsupportedOperationException` — even though no filtering is needed. The catch message "Physical Parquet row ranges require an Offset Index" is misleading for the same reason. There is no non-filtered fallback in this module; suggest calling `fileReader.readRowGroup(rowGroup.index)` when the selection covers the entire row group (a flag computed in nextRange(), where the absolute bounds are known).

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalReader.java
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    /**
+     * Opens a Parquet reader, reusing a remembered footer when the file is unchanged in size. The
+     * returned reader owns the stream opened here and closes it.
+     */
+    private ParquetFileReader open(URI uri, ParquetInputFile file) throws IOException {
+        ParquetReadOptions options = ParquetReadOptions.builder().build();
+        long size = file.getLength();
+        ParquetMetadata remembered = footers.get(uri, size);
```
SemyonSinchenko: The footer cache keys entries only by (uri, file size), so a file overwritten in place with different content of the same length is parsed with a stale footer — wrong schema, wrong row-group offsets, corrupt reads. The "chunk files are immutable" javadoc assumption is not enforced: the storage API exposes `OutputFile.createOrOverwrite()` (storage-api, storage-local, storage-s3) and this module's own writer accepts `WriteMode.OVERWRITE`. Consider invalidating cache entries on write paths, keying on a stronger signal, or at minimum documenting the immutability contract where the cache is defined.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
```diff
@@ -0,0 +1,350 @@
...
+        this.rowGroups = rowGroups(fileReader.getRowGroups());
+    }
+
+    @Override
+    public boolean next() throws IOException {
+        if (closed || exhausted) {
+            current = null;
+            return false;
+        }
+        if (emitted == limit) {
```
SemyonSinchenko: Two issues: (1) for limit > 0 this branch is unreachable — reaching the limit inside the batch loop already sets exhausted = true, so the `closed || exhausted` check above returns false first; the only reachable case is limit == 0. (2) When the limit is reached mid-scan, next() sets exhausted without closing the ParquetFileReader or any open pages (unlike natural exhaustion, which goes through finish()), so after next() returns false the cursor is in different release states depending on the exit path; a caller that stops at the first false return without calling close() leaks the file handle. Suggest closing the reader when the limit is reached while keeping `current` retrievable so batch() stays valid after the final true return.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+        int count = list.getFieldRepetitionCount(0);
+        if (count == 0) {
+            return List.of();
+        }
+        List<Object> values = new ArrayList<>(count);
+        org.apache.graphar.io.ColumnType element =
+                column.field().type().elementType().orElseThrow();
+        for (int elementIndex = 0; elementIndex < count; elementIndex++) {
+            Group elementGroup = list.getGroup(0, elementIndex);
```
SemyonSinchenko: The variable names are inverted relative to the Parquet spec terminology, which makes the element-extraction logic hard to verify: `list` holds the LIST group instance (the outer group), while `elementGroup` actually holds one instance of the middle repeated "list" wrapper group, not the element itself. The logic reads correctly (each wrapper instance contains one primitive element), but the naming suggests the primitive element is being treated as a group.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetOutputFile.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        @Override
+        public long getPos() throws IOException {
+            return output.position();
+        }
+
+        @Override
+        public void write(int value) throws IOException {
+            output.write(new byte[] {(byte) value});
```
SemyonSinchenko: This allocates a fresh one-byte array on every single-byte write. Parquet serializes many small values (page headers, variant lengths, footer metadata) through write(int), producing avoidable garbage per file. A reusable one-byte buffer is safe here because the writer stream is used by a single thread.

maven-projects/io-parquet/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+    <artifactId>graphar-io-parquet</artifactId>
+    <packaging>jar</packaging>
+    <version>${graphar.version}</version>
+
+    <name>graphar-io-parquet</name>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <hadoop.version>3.3.0</hadoop.version>
```
SemyonSinchenko: The latest is 3.5.0 (https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common/3.5.0), what was the reason of pinning 3.3.0?

maven-projects/io-api/src/main/java/org/apache/graphar/io/ObjectValueVector.java
keksmd: Fixed in eb0ad12. List values are copied with a null-tolerant unmodifiable copy; a null element is accepted when the element field is nullable and rejected with a message naming the column and index otherwise.

maven-projects/io-api/src/main/java/org/apache/graphar/io/ObjectValueVector.java
keksmd: Fixed in eb0ad12. List values are snapshotted recursively, so nested lists are unmodifiable copies at every depth (`snapshotsNestedListsAtEveryDepth`).

maven-projects/io-api/src/main/java/org/apache/graphar/io/RecordBatches.java
keksmd: Fixed in eb0ad12. `ObjectValueVector` now checks every value at construction: null in a non-nullable field, Java type per kind (documented in the class javadoc), integer range for INT8/16/32, and list elements recursively against the element field. Both `RecordBatches` factories go through it and document the `IllegalArgumentException`.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalReader.java
keksmd: Fixed in eb0ad12, reader and writer together. The writer emits the 3-level layout with a REQUIRED or OPTIONAL `element` following the element field's nullability; the reader accepts REQUIRED/OPTIONAL elements and maps an absent element to null. Tests: a file written directly with a Spark-style schema (`optional int64 element`, with a null entry) reads back, and our own files are asserted to carry the spec layout.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalWriter.java
keksmd: Changed together with the reader, see the reply there.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
keksmd: Changed in eb0ad12: a range that covers the whole row group now calls `readRowGroup(index)`, and only a partial range goes through `readFilteredRowGroup`; the error message now says a partial range needs the Offset Index. For the record, parquet-java 1.18 already short-circuits `readFilteredRowGroup` to `internalReadRowGroup` when the ranges cover the block, so plain scans did not fail in practice, but relying on that was implicit.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetPhysicalReader.java
keksmd: Fixed in eb0ad12. The cache is now opt-in: `new ParquetPhysicalReader(storage)` parses the footer every time, and the capacity constructor documents that the caller guarantees no file is rewritten in place while the reader is in use. Storage exposes no version or ETag signal today, so a stronger key would have to come from a storage-api change; I kept that out of this PR.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
keksmd: Fixed in eb0ad12. Reaching the limit now closes the file reader immediately while keeping the last batch retrievable; the top-of-method check only handles `limit == 0`. Test `releasesTheFileAsSoonAsTheLimitIsReached` asserts every opened stream is closed right after the final `next()` returning true.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetBatchCursor.java
keksmd: Renamed in eb0ad12: `listGroup` for the LIST group, `repeatedGroup` for each repeated `list` wrapper, `elementType` for the element.

maven-projects/io-parquet/src/main/java/org/apache/graphar/io/parquet/ParquetOutputFile.java
keksmd: Fixed in eb0ad12: a per-stream one-byte buffer is reused. Both current `PositionOutput` implementations copy the bytes synchronously, so reuse is safe.

maven-projects/io-parquet/pom.xml
keksmd: 3.3.0 was the Hadoop version parquet-java 1.18 itself compiles against (`provided` in its pom). Hadoop 3.5.0 is compiled for Java 17 (class file major 61), which the SDK's Java 11 baseline cannot load. In eb0ad12 I replaced `hadoop-common` + `hadoop-mapreduce-client-core` with the shaded `hadoop-client-api` + `hadoop-client-runtime` 3.4.2 (latest Java 8/11 line): io-parquet's runtime classpath drops from 135 jars to 19, and Hadoop's own third-party dependencies are relocated, so they no longer conflict with consumers.

